PR #8, opened 2015-04-03
Use eclipse plugin for submodules only

msteiger:
eclipse is not a big fan of nested projects :-(

This avoids creating an empty eclipse project for the root project.


immortius:
I'll take your word for it. :)

PR #13, opened 2015-04-18
Permissive sandbox switch

OvermindDL1:
Adding permissive sandbox support in to gestalt, does nothing that it did not before unless the appropriate switches are set (passing true into both of the new constructors for the ModuleEnvironment and ModuleSecurityManager classes).


immortius:
Hmm, I might have a poke and see if there is a way to rework things so that this can be supported without having to introduce a flag (and thus allow more complex extensions in the future). Otherwise this looks good.

gestalt-module/src/main/java/org/terasology/module/sandbox/ModuleClassLoader.java
```diff
@@ -123,6 +126,9 @@ ClassLoader getBaseClassLoader() {
         if (parentLoader != this && !(parentLoader instanceof ModuleClassLoader)) {
             if (permissionProvider.isPermitted(clazz)) {
                 return clazz;
+            } else if (isPermissive) {
```
immortius: Couldn't the permission check and log be handled by the permission provider?

immortius:
Ok, I ended up decoupling the PermissionProviderFactory from the ModuleSecurityManager (that had been bugging me for a bit), and I feel the end result is cleaner and will allow for alternate permission providers to be created easily. I added a wrapper factory for the permission check logging.

PR #17, opened 2015-05-01
Allow only dots in Version patterns

msteiger:
This PR restricts the separators between the individual parts of `Version` to dots.


immortius:
Good catch. Probably was intended in the first place.

PR #18, opened 2015-05-05
Added support for optional modules to DependencyResolver.

MarcinSc:


gestalt-module/src/main/java/org/terasology/module/DependencyInfo.java
```diff
@@ -37,6 +38,15 @@ public DependencyInfo(DependencyInfo other) {
         this.id = other.id;
         this.minVersion = other.minVersion;
         this.maxVersion = other.maxVersion;
+        this.optional = other.optional;
+    }
+
+    public boolean isOptional() {
```
immortius: Needs Javadoc

gestalt-module/src/main/java/org/terasology/module/DependencyResolver.java
```diff
@@ -68,7 +68,17 @@ public DependencyResolver(ModuleRegistry registry) {
      * @return A set of compatible modules based on the required modules.
      */
     public ResolutionResult resolve(Name rootModule, Name... additionalModules) {
-        return resolve(Varargs.combineToSet(rootModule, additionalModules));
+        return resolve(true, rootModule, additionalModules);
+    }
+
+    /**
+     * @param includeOptional   If optional modules should be included
```
immortius: What circumstances would a called include or not include optional modules?  How do they make that determination? Why not just have optional modules are included if possible?

immortius:
Please add a description for what this PR is adding and why. 

I'm curious what happens in the case that an optional dependency and a mandatory dependency have incompatible version spans.

MarcinSc:
I'll answer all concerns in one comment, rather than separately.

No, optional dependencies should only be included in the resolving when asked to, or are accessible via other non-optional path. The use case for it already exists in Terasology (which is why this PR was created).

Module "Fluid" has an optional dependency on "ModularComputers" and adds support for APIs for computers to be able to read and transfer fluids from fluid inventories. If someone wants to play with just "Fluid" module, they should not be forced to download and use "ModularComputers" module. However, if the "ModularComputers" module ends up in the mix somehow (via other dependency, or explicit choice) the APIs should be automatically accessible.

The reason for giving a choice of following optional dependencies is required, if we decide to switch gradle to use gestalt-module (as it should) to determine what to include in the project, as in that case - the project should just ignore the "optional" flag and always follow the dependency graph, as otherwise the code would not compile.

So, overall - optional flag matters only for selection of modules to play in game.

BTW, I found a bug in the existing code. The following test will fail, I might have to rewrite the resolver, keeping all tests and adding more to cover this issue and the optional ones:

    @Test
    public void shouldNotInvalidateResolvingIfVersionNotUsed() {
        ModuleRegistry registry = new TableModuleRegistry();
        Module core = createStubModule(registry, "core", "1.0.0");
        Module moduleA1 = createStubModule(registry, "a", "1.0.0");
        Module moduleA2 = createStubModule(registry, "a", "2.0.0");

        addDependency(core, "a");
        addDependency(moduleA1, "c");

        DependencyResolver resolver = new DependencyResolver(registry);
        ResolutionResult results = resolver.resolve(new Name("core"));
        assertTrue(results.isSuccess());
        assertEquals(Sets.newHashSet(core, moduleA2), results.getModules());
    }

MarcinSc:
Ok, I was able to write an implementation that satisfies all tests (including the new ones I've added) and is recursive on dependency depth, so should not cause stack overflow, unless someone creates a REALLY long chain of dependencies, but lets just ignore this case for now.

MarcinSc:
Ok, please review and comment or merge.

immortius:
I will review in more detail when I have time - I will also check whether
the bug could have been addressed without a rewrite so I can understand
what went wrong.

I would suggest a few things:
*  I didn't get from the code that the include optional parameter meant
force include them, as opposed to include vs ignore them. I think this
could be clearer by renaming to forceRequireOptionals or similar.
* Might be simpler to set that flag when constructing the resolver rather
than passing it through methods.
* Could have a specific optional constraint.
* It would be nice to have a more complete description of the behavior of
optional dependencies, particularly in the presence of conflicts. Should
optional dependencies be included if possible, or do they just add
constraints if they are included otherwise? I would have thought the user
would have to specifically ask for them - they don't just get pulled in if
available?

MarcinSc:
Hmm, the test seems to pass. I'm not sure how I saw it failing. Anyway I would have no clue on how to add the "optional" into it anyway.

immortius:
Is my understanding correct that an optional dependency should not cause the target module to be brought into the final set, but if the target module is brought in due to explicit request or a mandatory relationship the the version restrictions of the optional dependency will apply?

Any, as a first step I will be taking just the unit tests from this request.

MarcinSc:
That is correct, it's because of binary dependencies, and possibility of missing methods or signature changes. If a module is brought in, it has to match the optional requirements.

immortius:
I'm not talking about the force include mode, just general behavior (since you still haven't provided a description of what optional actually means).

MarcinSc:
I think I did in the comment with JUnit test. Optional dependencies are included in resolution, only if asked to (the boolean parameter) or if there is a non-optional path to them.

immortius:
I don't see any comments in the JUnit tests. Also the ```resolvesOptionalIfNotSpecified()``` test implies the opposite - that optionals are included even without a required path.

immortius:
Anyway, don't worry about it. I'm re-implementing optional support without the large scale dependency resolver changes, and allowing the choice OptionalResolutionStrategy between ```FORCE_INCLUDE```, ```INCLUDE_IF_AVAILABLE``` and ```INCLUDE_IF_REQUIRED```

MarcinSc:
No, in the comment (on this page) that has JUnit test in it.

The "resolvesOptionalIfNotSpecified", is a backward compatibility test. So if you don't specify if you want to include or not (the old code), they are getting included - the default.

immortius:
Hmm, don't agree with that. Optional dependencies didn't exist before, so there's nothing to be backwards compatible with.

MarcinSc:
But there (might) be usage of existing methods of Resolver. And all the existing cases should default to including the optional ones. If our Gradle idea task was using gestalt-module (as it should), the same for Jenkins, then the default for those should be to include the optional dependencies.

immortius:
I understand what you are getting at, but disagree still. The default behavior for optional should be what makes sense. And that is not treating optional dependencies as mandatory. That just makes it a trap for the majority of use cases. Having optional work correctly as default wouldn't break the idea task on release, so it is backwards compatible - but the task would need to be updated before optional modules came into use.

MarcinSc:
The general idea of backward compatibility default, is to have to update as little code as possible. If you introduce the default being "not include", then idea task, Jenkins, and everything else will have to be updated as well, at the same time, or will start falling.

If you however have the default as "test as required", then everything will continue working.

I'm fine either way, just wanted to avoid having to change every code that uses it.

immortius:
Hmm. But if it is "optional is mandatory" by default then the engine needs to be updated or it will break in the presence of optional. So I might as well go with the behavior that makes the most sense.

Either way it is backwards compatible in the sense that nothing breaks when you update to use the new version - it is only when you start to use the new feature that things need to be updated.

immortius:
Incidentally I've found that making a null version a tracked possibility by the dependency resolver goes a long way towards handling optional. I'm not sure the constraints have every been calculated correctly though, fleshing out some unit tests to check them more thoroughly.

PR #20, opened 2015-05-12
Addressing #18 - optional module support

immortius:
* Adds optional flag to DependencyInfo
 * Supports three strategies for handling optionals:
    * Treat as mandatory
    * Include if available
    * Include only if otherwise required
 * Reworked the dependency resolver to treat "no version" as a version possibility
 * Encapsulated the contraint processing in the constraint class


immortius:
@MarcinSc would appreciate your comments

MarcinSc:
It seems fine. I'm not sure if there will be any use for the "Include if available" mode.
Also, I think a few more negative tests might be nice.

PR #27, opened 2015-06-06
Add write support for ModuleMetadata

msteiger:
The first commits adds `equals` and `hashcode` for `ModuleMetadata`. This also includes the equality checks for  `I18nMap` and `DependencyInfo`.

The second commit adds write support for `ModuleMetadata` and renames the reader class to `ModuleMetadataJsonAdapter` as it also supports writing now. An alternative would be separate reader/writer classes, but this would also involve registering the same extensions to properly restore the data.

I also explicitly specified the default data format to ISO-8601 `yyyy-MM-dd'T'HH:mm:ss.SSSXXX"` (example: "2015-01-27T22:27:40.671-05:00") so that gson is able to work with `Date` instances without adapter classes. By default, gson just writes the date, but not the time, which isn't very useful.


immortius:
I... don't want to support writing ModuleMetadata?

gestalt-module/src/main/java/org/terasology/i18n/I18nMap.java
```diff
@@ -113,4 +113,25 @@ public String toString() {
     public Iterator<Map.Entry<Locale, String>> iterator() {
         return values.entrySet().iterator();
     }
+
+    @Override
+    public int hashCode() {
+        return 31 * values.hashCode();
```
immortius: Objects.hash() is preferred.

gestalt-module/src/main/java/org/terasology/i18n/I18nMap.java
```diff
@@ -113,4 +113,25 @@ public String toString() {
...
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
```
immortius: Use Objects.equals() to handle nulls concisely.

immortius: Generally I find this pattern works well:

```java
public boolean equals(Object obj) {
    if (obj == this) {
        return true;
    }
    if (obj instanceof TargetType) {
        TargetTypeother = (TargetType) obj;
        return Objects.equals(attribute1, other.attribute1) && Objects.equals(attribute2, other.attribute2);
    }
    return false;
}```

File: gestalt-module/src/main/java/org/terasology/module/ModuleMetadataJsonAdapter.java [hunk 3]
```diff
@@ -0,0 +1,214 @@
...
+    }.getType();
+
+    private final GsonBuilder builder;
+    private transient Gson cachedGson;
+    private final Map<String, Type> extensionMap = Maps.newHashMap();
+
+    public ModuleMetadataJsonAdapter() {
+        this.builder = new GsonBuilder()
+                .setPrettyPrinting()
+                .setDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX")
```
immortius: Is this the ISO 8601 format? Or would that end with Z?

immortius:
Otherwise I'm quite happy with the PR.

Cervator:
I'm feeling dense, but am not even sure what this is trying to do? :-)

At first I thought maybe it related to using JSON files for translation, but ...

gestalt-module/src/main/java/org/terasology/module/ModuleMetadataJsonAdapter.java
msteiger: Both are equally valid. The `Z` means zero offset  from UTC time. The `XXX` uses the (long format) timezone offset (e.g. -05:00). [Wikipedia has some details on the specs.](
http://en.wikipedia.org/wiki/ISO_8601#UTC). 
The `Z` variant silently assumes that `Date` is in UTC time, which is a potential source of error, imho.

msteiger:
It provides support to write `ModuleMetadata` (e.g. to `module.txt` files). Currently these files need to be created manually.

I'm currently using it in master-server to enrich existing `module.txt` files. They are read from jars in Artifactory, annotated with jar file size, checksum, timestamp and download link. 

In the next step, I want to create a `RemoteModule` that can be used for dependency resolution and replaced later by an `ArchiveModule` after downloading the actual jar file.

msteiger:
I addressed all comments and added read/write methods for `JsonReader` and `JsonWriter`, which are not compatible to `Reader` and `Writer`. This allows for reading/writing metadata as part of larger structures (e.g. lists).

gestalt-module/src/main/java/org/terasology/module/ModuleMetadata.java
```diff
@@ -179,4 +179,30 @@ public void setExtension(String extensionId, Object extension) {
...
+    public int hashCode() {
+        return Objects.hash(id, version, displayName, description, requiredPermissions, dependencies, extensions);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
```
immortius: No need for a null check here, instanceof would allow both the class and null check as one.

msteiger:
I updated the two other methods.

PR #34, opened 2015-06-27
Support explicit module versions

### LINKED ISSUE #33: Support resolving a particular module version
Currently, only module names can be used to resolve modules. When multiple versions of a module are available, it might be useful to resolve a particular version (or a range?) of a module.
For example, someone might want to play v0.9 even if v1.0 is already available.

The master-server uses a workaround based on a fake module to resolve a specified version of a module:

```java
ModuleMetadata meta = new ModuleMetadata();
Name fakeName = new Name("fake");
meta.setId(fakeName);
meta.setVersion(new Version(1, 0, 0));
DependencyInfo di = new DependencyInfo();
di.setId(name);
di.setMinVersion(version);
di.setMaxVersion(version.getNextPatchVersion());
meta.getDependencies().add(di);
DummyModule fakeMod = new DummyModule(meta);

moduleRegistry.add(fakeMod);
ResolutionResult result = dependencyResolver.resolve(fakeName);
moduleRegistry.remove(fakeMod);
```

I would find it useful if this was supported by DependencyResolver.

msteiger:
This PR enhances `DependencyResolver` to support specifying explicit module versions through a builder pattern.

I made this an inner class of `DependencyResolver`, which forwards the actual resolution process to the package private class `ResolutionAttempt`. These changes are backward compatible and tested by two unit tests.

Unfortunately, GitHub doesn't show move operations properly: I only added a version check to `populateDomains()` ([see line 94-95](https://github.com/MovingBlocks/gestalt/compare/develop...msteiger:dep-resolver?expand=1#diff-1915fdce829f21706458e0a9394c1201R94)) to filter the initial set of modules based on version ranges.

Fixes #33.

### DIFF
--- a/gestalt-module/src/main/java/org/terasology/module/DependencyResolver.java
+++ b/gestalt-module/src/main/java/org/terasology/module/DependencyResolver.java
@@ -64,7 +64,7 @@ public class DependencyResolver {
      * @return A set of compatible modules based on the required modules.
      */
     public ResolutionResult resolve(Name rootModule, Name... additionalModules) {
-        return builder().add(rootModule).addAll(additionalModules).build();
+        return builder().require(rootModule).requireAll(additionalModules).build();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


immortius:
Looks good. Can you add some javadoc around the builder class and its operation, and possibly improve the method names (I would argue it isn't very clear what adding a module does - it would be clearer with names like ```require``` and ```restrictToRange```, or similar).

msteiger:
Makes sense. I changed the method names and added JavaDoc where I saw need.

immortius:
Cheers!

PR #38, opened 2015-07-25
Make ModuleLoader use the new json adapter

msteiger:


immortius:
This is technically an API breaking change, so not sure I want to do this until 2.0.0. Assigning milestone as such, but leaving it for now.

msteiger:
The old, deprecated `ModuleMetadataReader` extends the new `ModuleMetadataJsonAdapter` class. Thus, existing code should continue to work.

PR #41, opened 2015-08-19
Fix AssetDataFile.toString()

msteiger:
Currently, `AssetDataFile.toString()` attempts to convert the path to an URI. This does not always work, in particular not for [ModulePath.toUri()](https://github.com/MovingBlocks/gestalt/blob/develop/gestalt-module/src/main/java/org/terasology/module/filesystem/ModulePath.java#L274), which does not implement the said method.

I therefore suggest using the absolute path to avoid this issue.

### DIFF
--- a/gestalt-asset-core/src/main/java/org/terasology/assets/format/AssetDataFile.java
+++ b/gestalt-asset-core/src/main/java/org/terasology/assets/format/AssetDataFile.java
@@ -106,7 +106,7 @@ public class AssetDataFile {
 
     @Override
     public String toString() {
-        return path.toUri().toString();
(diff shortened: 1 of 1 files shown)


immortius:
Does absolute path return a sensible value for assets in archives/zip files?

msteiger:
As far as I can tell, yes. An alternative would be returning the path as-is.

immortius:
I'm pretty sure what I was going for was having a mechanism that would contain the path to zip + internal path for archives, and just the path for normal files. But I removed uri support when I changed from per module to all module file system, and must have missed fixing this.

Your fix is good, merging.

PR #42, opened 2015-09-18
Improved asset disposal on asset garbage collection

immortius:
Switched to a PhantomReference based method for disposal of garbage collected assets.

This avoids the need for multiple garbage collection to fully collect assets.
With this change the Asset::onDispose() method is removed. Assets are now expected to register a static Runnable implementation with a provided DisposalHook, which will be run either when an asset is explicitly disposed, or if it is disposed after garbage collection via AssetType::processDisposal().


Cervator:
Any chance this or other related changes could help fix https://github.com/MovingBlocks/Terasology/issues/1855 more thoroughly? It appears to still be active in more rare cases, there is a user report at http://forum.terasology.org/threads/delta-too-great.1374 (lists a few issues, one matches the log snippet from 1855)

PR #55, opened 2016-05-07
Add ModuleAwareATM.disposeAll

msteiger:
Required by https://github.com/MovingBlocks/Terasology/pull/2306


immortius:
Hmm, can't you just get the assetType and then call disposeAll on that?

msteiger:
Yes, that works, too. Maybe the benefit of this new method is too marginal to be worth it.

PR #58, opened 2017-01-28
Asset path caching

MaxBorsch:
Speeds up asset loading (especially on consecutive loads, like rejoining a world)
Caches the paths to assets per module, so when the same module assets need to be reloaded the file tree is not scanned again. Limited to ~128 modules in cache so that there aren't memory problems if you reload worlds that have ~80 different modules each.

Also adds
`assetTypeManager.setUseAssetCaching(true);`
and
`assetTypeManager.setAutoReloadAssets(true); // Need setting to toggle this off for non-devs`
<hr></hr>
I tested both in-game and on a custom test, you can see the boost by adding something like:

`long dt = System.currentTimeMillis();`
        `assetTypeManager.setUseAssetCaching(true);//try false`
        `assetTypeManager.setAutoReloadAssets(true);`
        `assetTypeManager.switchEnvironment(moduleManager.getEnvironment());`
        `logger.info("ENVSwitch: " + (System.currentTimeMillis() - dt));`

to the `handleSwitchToGameEnvironment()` in `EnvironmentSwitchHandler.java`, then join and rejoin a world. With around 10 modules, loading consistently took ~14-15 seconds without caching, ~11-12 with caching.

### DIFF
--- a/gestalt-asset-core/src/main/java/org/terasology/assets/module/ModuleAssetDataProducer.java
+++ b/gestalt-asset-core/src/main/java/org/terasology/assets/module/ModuleAssetDataProducer.java
@@ -130,6 +130,11 @@ public class ModuleAssetDataProducer<U extends AssetData> implements AssetDataPr
     private final SetMultimap<Name, Name> resolutionMap = Multimaps.synchronizedSetMultimap(HashMultimap.<Name, Name>create());
 
     /**
+     * Used as the key to access the right set of assets from cache.
+     */
+    private final int assetCacheKey;
+
+    /**
      * Creates a ModuleAssetDataProducer
      *
      * @param moduleEnvironment   The module environment to load asset data from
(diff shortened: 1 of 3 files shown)


immortius:
Just starting to look at this, a couple of initial thoughts:

- Would like to see this taking advantage of the existing reflections cache system rather than introducing a new one. This would also mean that there wouldn't need to be any file system scanning for packaged modules with a compile-time reflections cache. This does require adding ```new ResourcesScanner()``` to the set of scanners used when building the reflections cache, and some method exposing these as module file system paths.

- I'm not a fan of the ```method(null)``` pattern. Would prefer no-parameter overloads for the null case - I like to provide an expectation that any method that has a parameter expects a non-null.

- The hash calculation logic should be on in the cache rather than outside of it, in my opinion - this prevents it being used differently by different users. Also it might be simpler to just cache everything under the assets/overrides/deltas, regardless of asset types, and allow asset types to be applied later - then you don't have to worry about the asset types involved.

immortius:
An update on this. I've started having a look through gestalt-asset-core, both with this in mind and other reworks to address inadequacies in the current architecture.

- I had a brief look at using the reflections cache like I suggest, and while it would work for archive modules, it would not work for path modules as their assets folder isn't in the classpath so is not scanned by Reflections. So that is out.

- Instead what I have done is pulled scanning for files completely out of ModuleAssetDataProducer into a separate class. This class can then introduce file caching - which I've done using guava's Cache classes. The discovered files are then fed to the ModuleAssetDataProducer through the same interface used by asset reloading.

- I've also split asset reloading out from ModuleAwareAssetTypeManager.

In general I'm looking to further decouple a lot of these elements.

immortius:
Closing this request. @MaxBorsch I didn't end up using your work directly, but it provided a good starting point for the caching that I ended up including, so thank you.

Issue #62, opened 2017-06-25
Instanced assets can not be used by modules


flo:
- AssetType#reload calls loadAsset within a  AccessController.doPrivilege block
- AssetType#createInstance does not call loadAsset within a AccessController.doPrivileged block

flo:
The problem has been observed with a screen

Cervator:
Closing as completed by #64 (released in v5.1.3)

PR #64, opened 2017-07-29
Allow modules to load instance assets.

flo:
This change probably fixes the instanced asset creation from issue #62 

I am not sure how to configure my terasology to use the gestalt source version.

I also didn't find a proper 5.1 branch to make this PR to as Terasology still uses 5.1.2. The commit in this PR is based on the latest 5.1.3-SNAPSHOT commit.

It would be great, if you @immortius could explain me how you test a gestalt change with Terasology without building  new version. Ideally this should also be documented so I created an issue for this: https://github.com/MovingBlocks/Terasology/issues/3027

### DIFF
--- a/gestalt-asset-core/src/main/java/org/terasology/assets/AssetType.java
+++ b/gestalt-asset-core/src/main/java/org/terasology/assets/AssetType.java
@@ -325,13 +325,16 @@ public final class AssetType<T extends Asset<U>, U extends AssetData> implements
         Optional<? extends Asset<U>> result = asset.createCopy(asset.getUrn().getInstanceUrn());
         if (!result.isPresent()) {
             try {
-                for (AssetDataProducer<U> producer : producers) {
-                    Optional<U> data = producer.getAssetData(asset.getUrn());
-                    if (data.isPresent()) {
-                        return Optional.of(loadAsset(asset.getUrn().getInstanceUrn(), data.get()));
+                return AccessController.doPrivileged((PrivilegedExceptionAction<Optional<T>>) () -> {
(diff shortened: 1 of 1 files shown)


flo:
This fix is btw. needed by the Scenario module. It uses it to allow you to allow you to define "event -> actions" logic ingame without programming knowledge. For this actions you can define recursivly expressions for which it is necessary to display the same dialog multiple times.

The GSOc student @ThompsonTyler (alias cata) works on that module and @nihal111 plans to use it for his gsoc project too.

flo:
Ok was able to test and I can confirm that the PR indeed did fix the issue.

To test I build gestalt with "./gradlew install" and then changed in engine/build.gradle:
```
-    compile group: 'org.terasology', name: 'gestalt-module', version: '5.1.2'
-    compile group: 'org.terasology', name: 'gestalt-asset-core', version: '5.1.2'
+    compile group: 'org.terasology', name: 'gestalt-module', version: '5.1.3-SNAPSHOT'
+    compile group: 'org.terasology', name: 'gestalt-asset-core', version: '5.1.3-SNAPSHOT'
```

Cervator:
Branching stuff is all set now :-)

I might simply make a new Gestalt job in Jenkins purely to build a v5.1.3

immortius:
This is a reasonable change. One request - I would like the error handling to be in line with #reload - it should get the cause from the privileged exception.

Cervator:
So essentially replace the one logger line with something like this?

`logger.error("Failed to register asset '{}'", asset.getUrn().getInstanceUrn(), e.getCause());`

flo:
@immortius done

PR #69, opened 2019-08-05
Minor tweaks for Android-less workspaces, Jenkins, Artifactory... 

Cervator:
Plus a weird unit test quirk on Win10.

Goal of this was to aid in https://github.com/MovingBlocks/DestinationSol/pull/418 to let us build & publish Gestalt at its v7 SNAPSHOT level so it could be tested via binaries for DS. Needed to readd some Artifactory-related pieces and what not. I also made it so you could work with Gestalt without necessarily having a full Android setup, which would allow easier local dev + let us build in Jenkins right now _without_ the Android piece (a bit tricky to set up in CI - but it is coming!)

Separately I also built and manually published the reflections snapshot dependency so it can just be pulled out of Artifactory rather than be built locally first.

This isn't clean quite yet - pretty much threw stuff at the wall to see what would stick and yield published binaries in Artifactory from a local build, with credentials added to `gradle.properties`. Need to go back and take back out the cruft that isn't needed. Code analytics are live again but out of date and really need an overhaul, the launcher has gone to SpotBugs instead of FindBugs for instance and there's a new set of analytics plugins in Jenkins to play with.

I also managed to put it in not quite the right place as it went to http://artifactory.terasology.org/artifactory/webapp/#/artifacts/browse/tree/General/terasology-snapshot-local/gestalt rather than libs-snapshot. More stuff to clean up and fix before this is ready :-)

Builds on https://github.com/MovingBlocks/gestalt/pull/68 naturally, with thanks to @immortius for getting this rolling and to @BenjaminAmos for putting it to work on DS!

Edit: I also ran all the unit tests (as perhaps indicated by including a workaround for one on Windows) _including_ the full Android setup with a simulator, all appeared well! Finally got Android Studio set up

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,11 +10,17 @@ buildscript {
         google()
         jcenter()
         mavenLocal()
+        maven {
+            url "http://artifactory.terasology.org/artifactory/virtual-repo-live"
+        }
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:3.2.1'
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
+
+        // Artifactory plugin
+        classpath(group: 'org.jfrog.buildinfo', name: 'build-info-extractor-gradle', version: '4.7.3')
     }
 }
 
@@ -24,9 +30,8 @@ allprojects {
         jcenter()
         mavenLocal()
         maven {
-            url "http://artifactory.terasology.org/artifactory/repo"
+            url "http://artifactory.terasology.org/artifactory/virtual-repo-live"
         }
-
     }
 }
 
@@ -37,25 +42,14 @@ task clean(type: Delete) {
     delete rootProject.buildDir
 }
 
-/*configure(subprojects.findAll { !it.name.contains("testpack") && !it.name.startsWith("module") }) {
+configure(subprojects.findAll { !it.name.contains("testpack") && !it.name.startsWith("module") && !it.name.contains("android")}) {
+    // General Java configuration
+    apply plugin: 'java'
     apply plugin: 'checkstyle'
     apply plugin: 'pmd'
     apply plugin: 'findbugs'
-    apply plugin: 'maven'
-    apply plugin: 'java-library-distribution'
-    apply plugin: 'artifactory-publish'
+    apply plugin: 'com.jfrog.artifactory'
     apply plugin: 'maven-publish'
-
-    // Primary dependencies definition
-    dependencies {
-        compile group: 'org.slf4j', name: 'slf4j-api', version: '1.7.21'
-        compile group: 'com.google.guava', name: 'guava', version: '19.0'
-
-        // These dependencies are only needed for running tests
-        testCompile group: 'junit', name: 'junit', version: '4.12'
-        testCompile group: 'ch.qos.logback', name: 'logback-classic', version: '1.1.7'
-        testCompile group: 'org.mockito', name: 'mockito-core', version: '1.10.19'
-    }
 
     // Set the expected module Java level (can use a higher Java to run, but should not use features from a higher Java)
     sourceCompatibility = 1.8
@@ -63,10 +57,9 @@ task clean(type: Delete) {
 
     jar {
         manifest {
-            attributes("Implementation-Title": project.name,
-                    "Implementation-Version": version)
+            attributes("Implementation-Title": project.name, "Implementation-Version": version)
         }
-        from(['LICENSE', 'NOTICE'])
+        from(['LICENSE'])
     }
 
     task sourceJar(type: Jar) {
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -37,36 +42,24 @@ task clean(type: Delete) {
     delete rootProject.buildDir
 }
 
-/*configure(subprojects.findAll { !it.name.contains("testpack") && !it.name.startsWith("module") }) {
+configure(subprojects.findAll { !it.name.contains("testpack") && !it.name.startsWith("module") && !it.name.contains("android")}) {
```
immortius: I'm not eager to reactive this section - I'm trying to avoid having the top-level build.gradle apply settings into other projects because I feel it confuses things.

gestalt-android/build.gradle
```diff
@@ -51,3 +57,61 @@ dependencies {
     androidTestImplementation 'com.android.support.test:runner:1.0.2'
     androidTestImplementation 'com.android.support.test.espresso:espresso-core:3.0.2'
 }
+
+
+// Artifactory publishing details
+buildscript {
```
immortius: On the other hand repeating this big section is a little awkward.

gestalt-asset-core/src/test/java/org/terasology/assets/module/autoreload/ModuleEnvironmentWatcherTest.java
```diff
@@ -108,7 +108,17 @@ public void test() throws IOException {
         }
         changed = watcher.checkForChanges();
         assertTrue(changed.containsEntry(assetType, new ResourceUrn(module.getId(), new Name("test.txt"))));
-        FilesUtil.recursiveDelete(tempDirectory);
+        try {
+            FilesUtil.recursiveDelete(tempDirectory);
+        } catch (IOException e) {
+            // Observed on Windows 10 to fail - timing issue? Try again after a brief moment!
```
immortius: I haven't had this fail under Windows 10. It is a bit odd.... perhaps you have a virus scanner or something that is locking and scanning the directory after creation?

Probably would suggest we rework it into a small look with pause and max attempts, and build that into the method.

settings.gradle
```diff
@@ -1,2 +1,8 @@
 rootProject.name = 'gestalt'
-include 'gestalt-util', 'testpack:testpack-api', 'gestalt-module', 'testpack:moduleA', 'testpack:moduleB', 'testpack:moduleC', 'testpack:moduleD', 'gestalt-asset-core', 'gestalt-entity-system', 'gestalt-es-perf', 'gestalt-bifrost', 'gestalt-benchmark', 'gestalt-android', 'gestalt-android-testbed'
\ No newline at end of file
+include 'gestalt-util', 'testpack:testpack-api', 'gestalt-module', 'testpack:moduleA', 'testpack:moduleB', 'testpack:moduleC', 'testpack:moduleD', 'gestalt-asset-core', 'gestalt-entity-system', 'gestalt-es-perf', 'gestalt-bifrost', 'gestalt-benchmark'
+
+if (new File("local.properties").exists()) {
```
immortius: Fair.

build.gradle
```diff
@@ -37,36 +42,24 @@ task clean(type: Delete) {
...
+    // General Java configuration
+    apply plugin: 'java'
```
immortius: This plugin isn't really appropriate for any of the projects (they should all be android or java-library)

build.gradle
```diff
@@ -37,36 +42,24 @@ task clean(type: Delete) {
...
     sourceCompatibility = 1.8
     targetCompatibility = 1.8
 
     jar {
         manifest {
-            attributes("Implementation-Title": project.name,
-                    "Implementation-Version": version)
+            attributes("Implementation-Title": project.name, "Implementation-Version": version)
         }
-        from(['LICENSE', 'NOTICE'])
```
immortius: I probably should at a NOTICE file.

build.gradle
Cervator: Plenty fair 👍 I had a nicer setup in mind a la Terasology where you just include suitable template build files, but that would take a bit longer and wasn't needed just yet. Will clean this up for sure :-)

gestalt-android/build.gradle
Cervator: Yep! Will handle.

gestalt-asset-core/src/test/java/org/terasology/assets/module/autoreload/ModuleEnvironmentWatcherTest.java
Cervator: Quite possible. I also still haven't gotten the branch working at all on my Mac. Will keep at it and not merge till I can validate it on 2-3 different systems (I've got a second Win10 system I can test too)

build.gradle
Cervator: I might be able to get around that by restructuring the Artifactory setup some more. Our usual standard relies on a few of the tasks in the plain java plugin that the java-library doesn't have. But really that's more just me having yet to dig in deeper on how the java-library works, will do 👍

build.gradle
Cervator: Yeah not sure where it went if we had one, or if not why this was in here in the first place. Copy paste ftw maybe :-)

Cervator:
Closing as subsumed into #68 !

PR #76, opened 2019-12-15
Optimise ArchiveFileSource

NicholasBatesNZ:
Taking @BenjaminAmos's custom [OptimisedArchiveFileSource class](https://github.com/MovingBlocks/DestinationSol/blob/develop/engine/src/main/java/org/destinationsol/modules/OptimisedArchiveFileSource.java) and bringing it straight here so that we can remove it from Destination Sol. The reason that it's needed is that `ZipFile` is much more performant than `ZipInputStream`.
The game freezes up using the latter and seems to run fine on the former


immortius:
Easy merge. Thanks

PR #77, opened 2020-04-17
removed upper and lower for name

pollend:
This issue for terasology shows kind of a key problem with transforming the name of the asset in gestalt. https://github.com/MovingBlocks/Terasology/issues/3887 I think we should avoid transforming the name at all and just use the name that was provided by the asset. Any kind of additional transformation on the type seems like a good way to introducing bugs. for one case it might be uppercase but another case it could be lower. When systems intersect is where we will see kind of a mix of the two and working out if its one or the other seems like a major problem.

### DIFF
--- a/gestalt-module/src/main/java/org/terasology/gestalt/naming/Name.java
+++ b/gestalt-module/src/main/java/org/terasology/gestalt/naming/Name.java
@@ -51,7 +51,7 @@ public final class Name implements Comparable<Name> {
     /**
      * @return The original name when string is passed into name
      */
-    public String originalName() {
+    public String displayName() {
         return originalName;
     }
 
--- a/gestalt-module/src/main/java/org/terasology/gestalt/naming/Name.java
+++ b/gestalt-module/src/main/java/org/terasology/gestalt/naming/Name.java
@@ -39,24 +39,24 @@ public final class Name implements Comparable<Name> {
      */
     public static final Name EMPTY = new Name("");
 
-    private final String originalName;
+    private final String normalizedName;
 
     public Name(String name) {
         Preconditions.checkNotNull(name);
-        this.originalName = name;
+        this.normalizedName = name.toLowerCase(Locale.ENGLISH);
     }
 
     /**
      * @return Whether this name is empty (equivalent to an empty string)
      */
     public boolean isEmpty() {
-        return originalName.isEmpty();
+        return normalizedName.isEmpty();
     }
 
 
     @Override
     public int compareTo(Name o) {
-        return originalName.compareTo(o.originalName);
+        return normalizedName.compareTo(o.normalizedName);
     }
 
     @Override
@@ -66,18 +66,18 @@ public final class Name implements Comparable<Name> {
         }
         if (obj instanceof Name) {
             Name other = (Name) obj;
-            return originalName.equals(other.originalName);
+            return normalizedName.equals(other.normalizedName);
         }
         return false;
     }
 
     @Override
     public int hashCode() {
-        return originalName.hashCode();
+        return normalizedName.hashCode();
     }
 
     @Override
     public String toString() {
-        return originalName;
+        return normalizedName;
     }
 }
--- a/gestalt-module/src/main/java/org/terasology/gestalt/naming/Name.java
+++ b/gestalt-module/src/main/java/org/terasology/gestalt/naming/Name.java
@@ -23,7 +23,7 @@ import java.util.Locale;
 import javax.annotation.concurrent.Immutable;
 
 /**
- * A name is a normalised string used as an identifier. Primarily this means it is case insensitive.
+ * A name is a normalisedName string used as an identifier. Primarily this means it is case insensitive.
  * <p>
  * The original case-sensitive name is retained and available for display purposes, since it may use camel casing for readability.
  * </p><p>
@@ -39,20 +39,20 @@ public final class Name implements Comparable<Name> {
      */
     public static final Name EMPTY = new Name("");
 
-    private final String normalised;
+    private final String normalisedName;
     private final String originalName;
 
     public Name(String name) {
         Preconditions.checkNotNull(name);
         this.originalName = name;
-        this.normalised = name.toLowerCase(Locale.ENGLISH);
+        this.normalisedName = name.toLowerCase(Locale.ENGLISH);
     }
 
     /**
      * @return Whether this name is empty (equivalent to an empty string)
      */
     public boolean isEmpty() {
-        return normalised.isEmpty();
+        return normalisedName.isEmpty();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


gestalt-module/src/main/java/org/terasology/gestalt/naming/Name.java
```diff
@@ -39,39 +39,33 @@
...
     public Name(String name) {
         Preconditions.checkNotNull(name);
         this.originalName = name;
-        this.normalisedName = name.toLowerCase(Locale.ENGLISH);
+        this.normalizedName = name.toLowerCase(Locale.ENGLISH);
     }
 
     /**
-     * @return Whether this name is empty (equivalent to an empty string)
+     * @return The original name when string is passed into name
```
skaldarnar, line 52:
```suggestion
     * @return The original <b>case-sensitive</b> name when string is passed into name
```

gestalt-module/src/main/java/org/terasology/gestalt/naming/Name.java
```diff
@@ -39,39 +39,33 @@
...
-    public String toLowerCase() {
-        return normalisedName;
+    public boolean isEmpty() {
+        return normalizedName.isEmpty();
     }
 
-    /**
-     * @return The Name in uppercase consistent with Name equality (so two names that are equal will have the same uppercase)
-     */
-    public String toUpperCase() {
```
skaldarnar, line 68: as gestalt is a basic library, I think we should use deprecation here...

gestalt-module/src/main/java/org/terasology/gestalt/naming/Name.java
```diff
@@ -39,39 +39,33 @@
...
-        return normalisedName.isEmpty();
+    public String originalName() {
+        return originalName;
     }
 
     /**
-     * @return The Name in lowercase consistent with Name equality (so two names that are equal will have the same lowercase)
+     * @return Whether this name is empty (equivalent to an empty string)
      */
-    public String toLowerCase() {
```
skaldarnar, line 61: Same here, add deprecation flag.

gestalt-module/src/main/java/org/terasology/gestalt/naming/Name.java
```diff
@@ -39,39 +39,48 @@
...
     public boolean isEmpty() {
-        return normalisedName.isEmpty();
+        return normalizedName.isEmpty();
     }
 
     /**
      * @return The Name in lowercase consistent with Name equality (so two names that are equal will have the same lowercase)
+     * @deprecated
      */
     public String toLowerCase() {
```
skaldarnar, line 69: I think we should use both `@deprecated` in the Javadoc to explain the deprecation and name alternatives and `@Deprecated` on the method itself to mark it as deprecated for the compiler.

I'm not sure how to properly link to other methods in Javadoc, though.

```suggestion
     * @deprecated This is scheduled for removal in upcoming versions. 
     *             Use {@code toString} or {@code displayName} instead.
     *             Note that a Name should not be transformed to a String for further processing.
     */
    @Deprecated
    public String toLowerCase() {
```

gestalt-module/src/main/java/org/terasology/gestalt/naming/Name.java
```diff
@@ -39,39 +39,48 @@
...
-        return normalisedName;
+        return normalizedName;
     }
 
     /**
      * @return The Name in uppercase consistent with Name equality (so two names that are equal will have the same uppercase)
+     * @deprecated
      */
     public String toUpperCase() {
```
skaldarnar, line 77: Same here for this deprecation.

```suggestion
     * @deprecated This is scheduled for removal in upcoming versions. 
     *             Use {@code toString} or {@code displayName} instead.
     *             Note that a Name should not be transformed to a String for further processing.
     */
    @Deprecated
    public String toUpperCase() {
```

gestalt-module/src/main/java/org/terasology/gestalt/naming/Name.java
```diff
@@ -81,20 +90,18 @@ public boolean equals(Object obj) {
...
         return false;
     }
 
     @Override
     public int hashCode() {
-        return normalisedName.hashCode();
+        return normalizedName.hashCode();
     }
 
     @Override
```
skaldarnar, line 103:
```suggestion
    /**
      * @return The Name as string consistent with Name equality (so two names that are equal will have the same {@code toString})
      */
    @Override
```

immortius:
Mmmm. I'm not against removing (or deprecating) the toUpper and toLower methods, but I think we should leave toString alone - toString exists for display, introducing a new function for this isn't necessary. Also don't think it is worth changing the spelling of normalised to a perhaps slightly more correct spelling when it is only used internally.

pollend:
my main reasoning would be that toString should return the id of the object and that choosing to display it would be a more deliberate process. we probably do more comparisons using toString vs cases where we would display the name of the object. The default behavior would be to give back the normalized version because in most cases it would be used for a comparison right?

skaldarnar:
I have to agree with both of you (to some degree). `toString` _should_ be used for display, but unfortunately it isn't. Maybe @immortius is right and the main problem is that (basically all of) Terasology's code is using it wrong, but that's still no reason for changing the library.

However, I also agree with @pollend that it is a bit of bad design if it is too easy to use the class in a wrong way... `toString` is used implicitly in some cases, and most (new) Java programmers learn about that method, and whenever they try to compare a `Name` to a String they'll use the (case sensitive) one...

I'm wondering whether something like `Names.compare(name, string)` would help here, as it might be a bit more straightforward than `name.compareTo(new Name(string))`  🤷‍♂️

immortius:
We could also add Name::compareTo(String) if that would be more straight forwards.

gestalt-module/src/main/java/org/terasology/gestalt/naming/Name.java
```diff
@@ -39,62 +39,81 @@
...
     public int hashCode() {
-        return normalisedName.hashCode();
+        return normalised.hashCode();
     }
 
+    /**
+     * @return The Name as string consistent with Name equality (so two names that are equal will have the same {@code toString})
```
skaldarnar, line 113: this no longer holds with the change reverted 😟 Now, we should rather put a disclaimer that it is only intended for display purposes.
```suggestion
     * @return The Name as string for display purposes
```

gestalt-module/src/main/java/org/terasology/gestalt/naming/Name.java
```diff
@@ -39,62 +39,81 @@
      */
     public static final Name EMPTY = new Name("");
 
+    private final String normalised;
```
skaldarnar, line 42: Let's keep it as it was - less noise in the diff.

```suggestion
    private final String normalisedName;
```

PR #78, opened 2020-05-14
Bumped gradle version to 5.6.4

casals:
Can't bump to 6.X since 'findbugs' is deprecated after 5.6.4

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -135,7 +135,7 @@ configure(subprojects.findAll { !it.name.contains("testpack") && !it.name.starts
 
 // gradle wrapper version
 wrapper {
-    gradleVersion '1.10'
+    gradleVersion '5.6.4'
 }
 
 ext {
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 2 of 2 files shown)


keturn:
Would [SpotBugs](https://spotbugs.github.io/) be an acceptable substitute for findbugs? that could be another issue, though.

casals:
> 
> 
> Would [SpotBugs](https://spotbugs.github.io/) be an acceptable substitute for findbugs? that could be another issue, though.

Actually, yes - findbugs is deprecated in gradle 6.x . I tried replacing it at first, but it also required a different config file, more changes in the script, etc - so I chose to bump the gradle version to the last one supporting findbugs for now (which solves the main problem of gradle being "too old").

skaldarnar:
Does any of you know how the back-and-forth porting is done here in _gestalt_ - I think the change also makes sense for `develop`.

keturn:
even with this branch, I get an error about not being able to find `findbugs`. Makes me doubt it's worth holding back on the gradle version in hopes of supporting it.

```
A problem occurred evaluating project ':gestalt'.
> Plugin with id 'findbugs' not found.
```

casals:
Actually `FindBugs` hasn't been developed since 2015, so it's more a matter of replacing it with `SpotBugs` (which uses a different syntax for configuration) or abandoning it at all. Nevertheless, `FindBugs` is working here with Gradle 5.6.4 - I double-checked the Gradle user guide and the plugin is supported up to version 5.6.x: https://docs.gradle.org/current/userguide/upgrading_version_5.html - perhaps triple-checking it with a fresh install/someone else + changing ti to `develop` as well (as skald suggested)

Cervator:
I did the things :-) #84 takes this a step further

casals:
Yay!! :D

PR #79, opened 2020-06-10
Added an EmptyComponent for components with no data

IsaacLic:


immortius:
I need a little convincing that this needs to be included in gestalt - as opposed to in whatever is using gestalt. EmptyComponents are outside of the norm, in my mind - I struggle to think of good use cases for them. But then I'm not familiar with how they're currently being used in Terasology.

Cervator:
I incrementally answered this to some degree on Discord, but wanted to add it in here for reference :-)

Both Terasology and DestinationSol are ending up with an increasing amount of marker components, just there to help filter event handlers. With TS nothing special is needed, since base components don't need to contain anything at all - but with DS using latest Gestalt you suddenly have to implement a dummy `copy` method for otherwise empty components, which does seem like it gets a little silly. Of course it'd be easy to then abuse that - if a marker component at some point gains data it shouldn't be allowed to extend `EmptyComponent` anymore. Maybe we can catch that with some sort of rule.

As for an example the first one I found for TS was in AdditionalRails: There's a given activator rail that when a cart passes over it any special function in the cart will be activated via `CartActivatedEvent` - so as a pattern in that module a bunch of different flavor carts just introduce an empty component via prefab to filter by and trigger its particular action when catching that event

gestalt-entity-system/src/main/java/org/terasology/gestalt/entitysystem/component/EmptyComponent.java
```diff
@@ -0,0 +1,16 @@
+package org.terasology.gestalt.entitysystem.component;
+
+/**
+ * An abstract component to be extended if a component that has no information to copy, like components that are flags
+ * for properties.
+ * <p>
+ * This should only be used for components that contain no information. If this is used for a component that does
+ * contain data, that data could be lost.
+ */
+public abstract class EmptyComponent implements Component<EmptyComponent> {
```
immortius, line 10: I think it might be nicer to keep the generic on EmptyComponent partially bound (i.e.
```
public abstract class EmptyComponent<T> implements Component<T extends EmptyComponent>
```

This would ensure the bounds on the final component's copy method is correct.

gestalt-entity-system/src/main/java/org/terasology/gestalt/entitysystem/component/EmptyComponent.java
```diff
@@ -0,0 +1,16 @@
+package org.terasology.gestalt.entitysystem.component;
+
+/**
+ * An abstract component to be extended if a component that has no information to copy, like components that are flags
+ * for properties.
+ * <p>
+ * This should only be used for components that contain no information. If this is used for a component that does
+ * contain data, that data could be lost.
```
immortius, line 8: It is a bit worse than data could be lost. The uncopied information would never get saved into the component store, so the component will not work as expected at all.

PR #80, opened 2020-06-10
feat: Improve Semver support

skaldarnar:
This is an attempt to improve the compatibility with Semver v2 as
defined by https://semver.org/ by introducing [**jsemver**](https://github.com/AntiLaby/jsemver) as internal
representation for the Version class.

Note that these changes lead to slightly different behavior in the
increment methods. Previously, the respective version level would be
increased while setting all less important numbers (to the right) to
zero. For instance

```java
new Version("1.2.0-SNAPSHOT").getNextMinorVersion()   // 1.3.0
```

With basing the internal representation on jsemver this changed to

```java
new Version("1.2.0-SNAPSHO").getNextMinorVersion()    // 1.2.0
```

Similar different behavior can be observed for the other parts.

With jsemver as internal representation we could easily support more of
Semver specification in _gestalt_ by exposing both the full _prerelease_
information and the _build metadata_.
I'd also like to offer methods for version increments which will
directly set the prerelease information, like

```java
version.getNextMinorVersion("SNAPSHOT")
```

This would be helpful for implementing CLI tooling in Terasology to
easily prepare a specific release and switch to the next snapshot
version.

Please feel free to share suggestions or opinions on this. I was also
considering to implement the Semver specification from scratch or
whether we should fork the whole project...

### DIFF
--- a/gestalt-module/build.gradle
+++ b/gestalt-module/build.gradle
@@ -24,6 +24,12 @@
 apply plugin: 'java-library'
 apply plugin: 'maven-publish'
 
+repositories {
+    maven {
+        url = 'https://heisluft.tk/maven/'
+    }
+}
+
 // Primary dependencies definition
 dependencies {
     api 'org.terasology:reflections:0.9.12-MB'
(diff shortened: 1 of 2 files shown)


immortius:
Looks good, I would be happy to have further functionality exposed. I guess the state of the jsemver library is something we'll have to keep an eye on - as it hasn't had its first major version release yet, nor is it in maven central (also a problem for gestalt, which I would like to address one day).

skaldarnar:
>  [...] as it hasn't had its first major version release yet, nor is it in maven central

Agreed on both. In worst case we'll have to add _gestalt-semver_ and maintain it ourselves 🙈

PR #82, opened 2020-07-20
fix: Mark `testpack-api` as Java 8 compatible to avoid any dependency resolution errors

eviltak:


DarkWeird (approved):
Ok.

Yeah seems it can be problem, if you using Java 11 by default, then testapack use it too...
but another project explicitly using java 8 .

skaldarnar:
@DarkWeird @eviltak so... are we going to merge this? Who clicks that button? 🤓

PR #83, opened 2020-09-06
WIP: gestalt DI

pollend:


pollend:
I have a transaction object that tracks changes that will be committed after a successful injection. need to also have a stack that tracks for objects so I can exit out early if I find a dependency cycle. 

```
BeanContext b = new BeanContext(registry);
b.inject(target)
```

```
BeanContext b = new BeanContext(registry);
try(transaction = new BeanTransaction()) {
  b.inject(target, transaction);
  b.inject(target1, transaction);
  b.inject(target2, transaction);
}
```

pollend:
I finished up the DefaultBeanContext and that should be enough for this to basically work to some degree. I still need to write some test cases to verify what I expect.

PR #87, opened 2021-02-27
feat: implement dependency injection

pollend:
This is partially inspired by some of the patterns adopted with how the current Context Scheme works in Terasology with some inspiration from Dagger, Micronaut and guice.

The BeanContext is created with a Service Registry that defines all the dependencies linked to a given context. Resources request from the context will attempt to resolve dependencies from the current context and then tries to look at a parent context if the dependency is not resolved there. 

BeanEnvironment is an internal cache that is used to track loaded in dependencies that are defined from the annotationProcessor. this is loaded in through a META-INF/services config that is generated by classes marked with `@Inject`/`@Introspected`/`@Singleton`/`@Transient`. the class definitions are fully defined so there is a minimum amount of scanning over the application classpath. This dramatically improves application boot times. 

`@Singleton`/`@Transient`  are not drive by anything without defining a scanner that can be used to search for dependencies that match a given annotation or pattern.   A sample is defined in org.terasology.gestalt.di.scanners but more standard ones can be added for convince. 

Dependencies are associated with 3 different scopes
- Singleton - lives as long as the context is still active
- Transient - new instance for every request or injection
- Scoped - new instance for every layer

Usage

```
Class DepA {
   @Inject
   protected DepB;
}

Class DepB {
}


static void main() {
   ServiceRegistry registry = new ServiceRegistry();
   registry.with(DepA.class).lifetime(Lifetime.Singleton);
   registry.with(DepB.class).lifetime(Lifetime.Singleton);

  BeanContext cntx = new DefaultBeanContext(registry);
  DepA  a = cntx.get(DepA.class);
}

```

features:

- [ ] keep a stack chain of dependencies and print out a tree with how a dependency failed to resolve
- [ ] catch cyclic dependencies and throw an exception where depA --> DepB --> DepA
- [ ] support Lazy<T> type dependencies that can be resolved at a later point.
- [x] support Optional<T> type dependencies
- [x] for service registry we should have suppliers with multiple inputs that can be resolved.
- [x] support Provider<T> type dependencies that can be resolved at request.
- [x] replaces `Reflections` with own `ClassIndexProcessor` and `ResourcesProcessor`  which works with gradle and idea in same way.
- [x] integrated `gestalt-module` you can use DI inside modules.


Cervator:
Just as an FYI to any interested parties: This PR as of right now is available as `7.0.6.di-SNAPSHOT` via our Artifactory, although I had to hack out all the `testpack` stuff over some weirdness with annotation processing. Naturally that would have to be fixed if we move forward with this approach 👍

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/BeanDefinitionProcessor.java
```diff
@@ -0,0 +1,398 @@
...
+
+
+    private static final String[] TARGET_ANNOTATIONS = new String[]{
+            "javax.inject.Inject",
+            "javax.inject.Qualifier",
+            "javax.inject.Singleton",
+            INTROSPECTED_CLASS,
+            SCOPED_CLASS,
+            TRANSIENT_CLASS
+    };
```
pollend, line 66: target annotations that will generate bindings.

keturn:
Surface-level impressions:

In general, I worry when I hear "implement a new dependency injection framework" given all the existing implementations already available. But not being very in tune with the Java ecosystem myself, I am in no position to suggest them by name or judge if they're a good fit for our application.

I am reassured by the fact that you _have_ looked at some of these other implementations and learned from them. Some "why not dagger?" section would be a helpful thing to have in the documentation.

The other thing that jumps out to me as someone relatively new to Java is ***"Bean"***.

Partly because beans are funny,  
Beanie Sandfurbs with angel wings, captioned "Hope is the thing with beans"

Partly because I know that "bean" has long been a thing with Java but I'd never bothered figuring out what the term means. (Initially used because java -> coffee -> coffee beans?) Having now looked it up, it seems like "NetBeans" is an IDE, "Enterprise JavaBeans" is some specific thing and "Spring Beans" means something in the Spring framework, but ["Bean" itself](https://www.oracle.com/java/technologies/javase/javabeans-spec.html) means just a class that:

* has zero-argument constructor
* Implements [`Serializable`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/Serializable.html)
* has no `public` properties, only methods

(That last one is funny to me because this practice is so widespread in Java that I'd forgotten you _could_ have a publicly writable property.)

is that how the term is used here?

[more thoughts later after I've actually looked at any of the details]

pollend:
@keturn this is a more defined pattern that is based on our context scheme that we use. it makes it easier to adopt into the existing codebase then try migrating to an existing DI implementation. It takes some inspiration of Dagger and adopts a couple patterns from Micronaut. Dagger dependencies are baked in stone so it makes it kind of hard to extend this out to module space and Micronaut is built more for applications with a single lifecycle. Its kind of a franken combination of the two with upfront definitions of services similar to guice but you can also define a scanner that looks for objects that fit a criteria and add them to the service definition. we don't have to use beans it could be something entirely different but I think people have some understanding of Bean if they've used spring before.  

 Games are a bit unique in that they have many different stages to startup with things getting built up and torn down. Thats my reasons for building a custom DI scheme and i'm sure we can build out more features to have closer interrogation with gestalt.

gestalt-inject/src/main/java/org/terasology/context/SoftServiceLoader.java
```diff
@@ -0,0 +1,97 @@
...
+                        return (S) classLoader.loadClass(clazz).getDeclaredConstructor().newInstance();
+                    } catch (ClassNotFoundException e) {
+                        e.printStackTrace();
+                    } catch (IllegalAccessException e) {
+                        e.printStackTrace();
+                    } catch (InstantiationException e) {
+                        e.printStackTrace();
+                    } catch (NoSuchMethodException e) {
+                        e.printStackTrace();
+                    } catch (InvocationTargetException e) {
```
keturn, line 84: my worries about swallowing all these exceptions aside, this could be more concise with `(ClassNotFoundException | IllegalAccessException` etc

keturn (commented):
I found I was kinda zoning out by the time I got to BeanDefinitionProcessor. Let me know if there's anything there or later that's of particular interest for review and I can try another pass.

I commented on first (judging-the-book-by-its-cover) impressions, and this review contains inline comments on some of the line-by-line reading. I think I'm missing a sense of what it's actually like to _use_ this; that would come from examples. (Or maybe, to a lesser extent, JavaDoc with just the public-facing stuff included.)

Overall it gives the impression of being well-structured. I noticed there are test cases and I appreciated how concise each one is. I didn't spend enough time with them to get a sense of whether they provide enough test coverage or not.

I think that's all the reviewingness I have in me at the moment. Cheers!

keturn:
I saw a lot of "(c) the original authors 2017 - 2020" stuff in `org.terasology.context`, was that moved here from elsewhere?

gestalt-di/src/main/java/org/terasology/gestalt/di/BeanContext.java
```diff
@@ -0,0 +1,35 @@
...
+
+    <T> List<T> getBeans(Class<T> clazz, Qualifier qualifier);
+
+    <T> Optional<T> findBean(BeanKey<T> identifier);
+
+    <T> Optional<T> findBean(Class<T> clazz);
+
+    <T> Optional<T> findBean(Class<T> clazz, Qualifier qualifier);
+
+    BeanContext getNestedContainer();
```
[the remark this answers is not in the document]
pollend: creates a new context that is another scope down from this. This new context parent is this context.

gestalt-di/src/main/java/org/terasology/gestalt/di/BeanEnvironment.java
```diff
@@ -0,0 +1,275 @@
...
+        }
+
+        if (Strings.isNullOrEmpty(prefix)) {
+            return Optional.of(Range.closed(0, input.length - 1));
+        }
+
+        int startPoint = 0;
+        int endpoint = input.length - 1;
+        while (true) {
+            String stringToTest = input[startPoint + (endpoint - startPoint) / 2].prefix;
```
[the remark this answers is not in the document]
pollend: yep, its doing binary search on namespace names. so we can filter them down.

gestalt-di/src/main/java/org/terasology/gestalt/di/BeanEnvironment.java
```diff
@@ -0,0 +1,275 @@
...
+            };
+        }
+        return Collections::emptyIterator;
+    }
+
+    public Iterable<BeanDefinition<?>> byPrefix(String prefix) {
+        return Iterables.concat(beanLookup.keySet().stream().map(k -> byPrefix(k, prefix)).collect(Collectors.toList()));
+    }
+
+    public <T> Iterable<BeanDefinition<?>> byPrefix(ClassLoader loader, String prefix) {
```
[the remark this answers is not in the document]
pollend: should be a BeanDefintion<T> so probably a mistake.

pollend:
You can look here for how this is used in practice. https://github.com/movingblocks/destinationsol/pull/586

gestalt-di/src/main/java/org/terasology/gestalt/di/BeanEnvironment.java
```diff
@@ -0,0 +1,275 @@
...
+    public <T> Iterable<BeanDefinition<?  extends T>> byInterface(ClassLoader loader, Class<T> targetInterface) {
+        final ClassLookup lookup = beanLookup.get(loader);
+        if (!lookup.interfaceIndex.containsKey(targetInterface)) {
+            return Collections::emptyIterator;
+        }
+        return Arrays.stream(lookup.interfaceIndex.get(targetInterface)).map(k -> (BeanDefinition<? extends T>) k.definition).collect(Collectors.toList());
+    }
+
+    public <T> Iterable<BeanDefinition<? extends T>> byInterface(Class<T> targetInterface) {
+        return Iterables.concat(beanLookup.keySet().stream().map(k -> byInterface(k, targetInterface)).collect(Collectors.toList()));
```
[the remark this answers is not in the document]
pollend: I concat the results of the individual classloaders. This does a lot of the similar stuff we are doing with reflections.

gestalt-di/src/main/java/org/terasology/gestalt/di/Lifetime.java
```diff
@@ -2,6 +2,15 @@
...
 package org.terasology.gestalt.di;
 
+/**
+ * lifetime scope drives how persistent a bean will and how it will be reused between different scopes. a child scope that redfined a bean will shadow its parents even if a
+ * parent scope has a singleton objects.
+ *
+ * - Scoped --> A single instance is defined for every nested scope. (A --> B --> C) a scoped instanced defined at B will create a new single instance in C and B
+ * - ScopedToChildren --> A new instance is defined for every nested scope except the one its defined in.   (A --> B --> C --> D) a scoped instanced defined at B will create a new single instance in C and D
+ * - Singleton  --> A single instance is defined and visible to all children.
+ * - Transient --> A new object is always returned if requested.
```
keturn, line 12: JavaDoc is more-or-less HTML, so check to see how this is rendered. I fear it might make a mess of this.

gestalt-inject/src/main/java/org/terasology/context/AnnotationMetadata.java
```diff
@@ -5,6 +5,9 @@
 import java.lang.annotation.Annotation;
 import java.util.List;
 
+/**
+ * Annotation data that is used to drive avoid reflection.
```
keturn: "drive avoid"?

keturn (commented):
hooray for docs

gestalt-di/src/main/java/org/terasology/gestalt/di/BeanEnvironment.java
```diff
@@ -0,0 +1,314 @@
...
+                break;
+            }
+            if (startPoint == endpoint) {
+                return Optional.empty();
+            }
+            if (stringToTest.compareTo(prefix) > 0) {
+                endpoint = startPoint + ((endpoint - startPoint) / 2);
+            }
+            if (stringToTest.compareTo(prefix) < 0) {
+                if (input[endpoint].prefix.compareTo(prefix) < 0) {
```
lyndseyBeil, line 111: I detect that this code is problematic. According to the [Correctness (CORRECTNESS)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#correctness-correctness), [SA: Self comparison of value with itself (SA_LOCAL_SELF_COMPARISON)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#sa-self-comparison-of-value-with-itself-sa-local-self-comparison).
This method compares a local variable with itself, and may indicate a typo or a logic error. Make sure that you are comparing the right things.

pollend:
I update master with develop and tagged it v7.0.7 so I think we should be good on that end just need to get approval for this PR @DarkWeird. anything we need to finalize since I think this is in a decent state to develop a v8.0.0 of this library.

DarkWeird (approved):

BenjaminAmos (approved):
Some minor comments here but the code works.

gestalt-inject-java/build.gradle
```diff
@@ -0,0 +1,97 @@
...
+ * - Gradle Documentation: http://gradle.org/documentation
+ * - View tasks for this project: $ gradlew tasks
+ */
+
+apply plugin: 'java-library'
+apply plugin: 'maven-publish'
+
+repositories {
+    maven {
+        url = 'https://heisluft.tk/maven/'
```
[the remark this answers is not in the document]
pollend: I talked with @Cervator and I think we should publish it to our maven.

gestalt-di/src/main/java/org/terasology/gestalt/di/index/UrlClassIndex.java
```diff
@@ -0,0 +1,125 @@
...
+ * <p>Can load index from:</p>
+ * <p>1. directory</p>
+ * <p>2. archive</p>
+ * <p>3. ClassLoader</p>
+ */
+public class UrlClassIndex implements ClassIndex {
+    private static final String METAINF = "META-INF";
+    // HACK: Android always claims that the URL of the "META-INF" directory or any of its subdirectories doesn't exist.
+    //       It will return the URL of any files within the "META-INF" directory and its subdirectories though.
+    private static final String METAINF_TEST_FILE = METAINF + "/gestalt-indexes-present";
```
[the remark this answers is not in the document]
pollend: umm, that would be a question for @DarkWeird.

gestalt-di/src/main/java/org/terasology/gestalt/di/BeanEnvironment.java
pollend: input[endpoint].prefix and prefix are different variables

lyndseyBeil: Noticed :) thank you for your kind reply

PR #89, opened 2021-03-06
feat: implement lambda injection

pollend:


soloturn:
can be closed, as this is now in gestalt-8 isn't it @BenjaminAmos ?

BenjaminAmos:
It is unclear about what this implements but I believe it was never completed. If I were to guess, I would say that this relates to passing beans from the environment as parameters to bean suppliers, which is functionality not present on `develop`.

Something like this (although this may not be valid syntax):
```java
serviceRegistry.with(ClassA.class).use((BeanA A, BeanB B) -> new ClassA(A, B));
```

This has significant conflicts and has done so for years, therefore I believe it can be safely closed.

BenjaminAmos:
I am going to leave the branch entact though as a reference for if somebody wishes to re-visit this.

PR #91, opened 2021-03-12
Integrate gestalt-di to gestalt-module

DarkWeird:
* removes `Reflections`
* `gestalt-di` provide indexes
* Provides 2 new AnnotationProcessor:
  * ResourceProcessor - creates index processor for fast lookup resources(assets etc) (required custom setup)
  * SubtypeProcessor - creates class index by annotation and subtypes
* Using Annotation processing for Class and Resource indexes. ( works same for idea and gradle, except detect changes at resources from IDEA)
* ResourceIndex using for ClassPathFileSource ( Directory and archive file sources is not changed)
* ClassIndex(SubtypesProcessor) using for `ModuleEnvironment#getSubtypes` and `ApiScanner` only. (compactability reason, part of classes will getting via `BeanContext`)
* Adapt API scanner( hehe packages with `@API` marks all classes within instead separate package handling)
* Integrate `BeanContext` to `ModuleEnvironment`

### DIFF
--- a/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/management/MapAssetTypeManager.java
+++ b/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/management/MapAssetTypeManager.java
@@ -17,17 +17,13 @@
 package org.terasology.gestalt.assets.management;
 
 import com.google.common.base.Preconditions;
-import com.google.common.base.Predicate;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.MapMaker;
 import com.google.common.collect.Multimaps;
-
 import net.jcip.annotations.ThreadSafe;
-
-import org.reflections.ReflectionUtils;
 import org.terasology.gestalt.assets.Asset;
 import org.terasology.gestalt.assets.AssetData;
 import org.terasology.gestalt.assets.AssetFactory;
@@ -35,7 +31,6 @@ import org.terasology.gestalt.assets.AssetType;
 
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -109,11 +104,11 @@ public final class MapAssetTypeManager implements AssetTypeManager {
         Preconditions.checkState(assetTypes.get(assetType.getAssetClass()) == null, "Asset type already registered for - " + assetType.getAssetClass().getSimpleName());
 
         assetTypes.put(assetType.getAssetClass(), assetType);
-        for (Class<?> parentType : ReflectionUtils.getAllSuperTypes(assetType.getAssetClass(), (Predicate<Class<?>>) input -> Asset.class.isAssignableFrom(input) && input != Asset.class)) {
-            subtypes.put((Class<? extends Asset>) parentType, assetType.getAssetClass());
-            (subtypes.get((Class<? extends Asset>) parentType)).sort(Comparator.comparing(Class::getSimpleName));
-
-        }
+//        for (Class<?> parentType : ReflectionUtils.getAllSuperTypes(assetType.getAssetClass(), (Predicate<Class<?>>) input -> Asset.class.isAssignableFrom(input) && input != Asset.class)) {
+//            subtypes.put((Class<? extends Asset>) parentType, assetType.getAssetClass());
+//            (subtypes.get((Class<? extends Asset>) parentType)).sort(Comparator.comparing(Class::getSimpleName));
+//
+//        }
     }
 
     /**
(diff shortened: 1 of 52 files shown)


keturn (commented):
thought: ResourceProcessor is going to be hard to gradle-optimize. I don't expect `compileJava` to normally consider the files in `resources` for its up-to-date checks.

I'm not too worried about the lack of incremental processing; walking the file tree and saving a flat list of its contents is going to be as fast as anything the compiler does.

I am worried about it being stale. It might not get re-run when you add or remove resources.

I can see the appeal of using an AnnotationProcessor: it's part of the existing tool chain, maven or gradle or IntelliJ or whatever, and doesn't require any configuration other than passing that one option. But at least for dev workspaces where people might be actively changing resources, we'll probably want this done in a gradle step instead.

That probably means that build environment just won't set that `resource` setting in the processing environment. Unless there's some use case that would want to allow both the ResourceProcessor _and_ the build tool to make lists and include both of them?

keturn:
> But at least for dev workspaces where people might be actively changing resources, we'll probably want this done in a gradle step instead.

**Or**, for dev workspaces, you skip this step entirely because there's a run-time fallback, right? At least on JVM, if not Android?

pollend:
one thing to note is that the namespaces names are repeated so we don't need to really store them as a flat list. There is probably a more optimal layout since the entries have a lot of repetition.  we should do more on the compiling side to limit what we need to accomplish during runtime. One thing to note is the namespace is repeated a lot and each entry is repeated so those strings could be stored an an index of sorts. if you sort the list alphabetically then each entry in your index can provide a offset into the array. so does this help with what we are doing with reflection to look up classes and instantiate them. I already kind of index them if they have the right annotation so they should be loaded into the BeanEnviroment. 

Something like this maybe :?.
```
[(A, 0, 0), (B,1,0), (C,1,6)
A.B
A.B
A.B
A.B
A.B
A.B
A.B.C
A.B.C
A.B.C
A.B.C
```

DarkWeird:
> one thing to note is that the namespaces names are repeated so we don't need to really store them as a flat list. There is probably a more optimal layout since the entries have a lot of repetition. we should do more on the compiling side to limit what we need to accomplish during runtime. One thing to note is the namespace is repeated a lot and each entry is repeated so those strings could be stored an an index of sorts. if you sort the list alphabetically then each entry in your index can provide a offset into the array. so does this help with what we are doing with reflection to look up classes and instantiate them. I already kind of index them if they have the right annotation so they should be loaded into the BeanEnviroment.
> 
> Something like this maybe :?.
> 
> ```
> [(A, 0, 0), (B,1,0), (C,1,6)
> A.B
> A.B
> A.B
> A.B
> A.B
> A.B
> A.B.C
> A.B.C
> A.B.C
> A.B.C
> ```

We can just to use `radix trie` - ideally. isn't it?

Where i can to find you index?

DarkWeird:
@keturn 
Resolve staling resources for gradle. (not works for `Intellij`)
Just sets resource to compileJava's input.

No any runtime Classpath's resource resolving currently.

keturn:
> Just sets resource to compileJava's input.

that should make sure it recognizes `compileJava` is not up-to-date when resources are changed, yeah.

is it smart enough to realize that it doesn't need to rebuild all the `.class` files if it's only the resources that have changed?

it _might_ be but that's deeper than I've looked at the internals of what happens during `compileJava`.

keturn:
> No any runtime Classpath's resource resolving currently.

hmm, so asset reloading only works for **re**loading assets with names that were present at compile time, can't add assets at runtime?

not familiar enough with the workflow to guess how much of a problem that is. Maybe it's fine because the usual pattern is making edits to an asset and seeing how they look. More of a problem if you were, say, trying out different textures and you wanted to _add_ and switch between them instead of overwriting.

DarkWeird:
> > No any runtime Classpath's resource resolving currently.
> 
> hmm, so asset reloading only works for **re**loading assets with names that were present at compile time, can't add assets at runtime?
> 
> not familiar enough with the workflow to guess how much of a problem that is. Maybe it's fine because the usual pattern is making edits to an asset and seeing how they look. More of a problem if you were, say, trying out different textures and you wanted to _add_ and switch between them instead of overwriting.

I forget that `ClasspathFileSource` stores files names in list. then yes. can't add assets at runtime.
Solutions:
1. Make `ClasspathFileSource` to re-reading index on each request. ( slower)
2. Make smart cache, which will be invalidate when index file will rewrote.
Then you can  add and remove assets during runtime then press (Сtrl+Shift+f9 for recompile). works at  gradle only.

It can problem for engine-dev only.

Modules using DirectoryFileSource and ArchiveFileSource (they not changed)

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/management/MapAssetTypeManager.java
```diff
@@ -52,6 +48,17 @@
     private final ListMultimap<Class<? extends Asset>, Class<? extends Asset>> subtypes =
             Multimaps.synchronizedListMultimap(ArrayListMultimap.<Class<? extends Asset>, Class<? extends Asset>>create());
 
+    private static Iterable<Class<?>> getAllSuperTypesBetween(Class<?> upperBound, Class<?> lowerBound) {
```
pollend, line 51: maybe something like this is more understandable: 
```suggestion
    private static Iterable<Class<?>> getAllSuperTypes(Class<?> from, Class<?> to) {
```

pollend:
umm would it hurt just to merge this with the DI branch and then try to resolve the complications in DestinationSol or would we do this is another phase?

DarkWeird:
> umm would it hurt just to merge this with the DI branch and then try to resolve the complications in DestinationSol or would we do this is another phase?

We can do this now.
Or... We can do it assets and entity system integration(there will two another service registry for them

PR #92, opened 2021-03-24
chore: replace jcenter with mavenCentral and gradlePluginPortal

pollend:
Locally building with ./gradlew build --no-build-cache --refresh-dependencies succeeded.
Adds to MovingBlocks/Terasology#4582

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,8 +8,9 @@
 buildscript {
     repositories {
         google()
-        jcenter()
-        mavenLocal()
+        mavenCentral()
+        // org.jetbrains.trove4j:trove4j
+        gradlePluginPortal()
         maven {
             url 'https://heisluft.de/maven'
         }
@@ -44,8 +45,9 @@ buildscript {
 allprojects {
     repositories {
         google()
-        jcenter()
-        mavenLocal()
+        mavenCentral()
+        // org.jetbrains.trove4j:trove4j
+        gradlePluginPortal()
         maven {
             url "http://artifactory.terasology.org/artifactory/libs-release-local"
         }


build.gradle
```diff
@@ -8,8 +8,9 @@
 buildscript {
     repositories {
         google()
-        jcenter()
-        mavenLocal()
+        mavenCentral()
+        // org.jetbrains.trove4j:trove4j
+        gradlePluginPortal()
```
BenjaminAmos, line 13: Just a note but this doesn't actually resolve `org.jetbrains.trove4j` from the gradle plugin portal. The gradle plugin portal fowards any unknown dependency requests to jcenter automatically, for historical reasons. This is mentioned in their recent post on the subject: https://blog.gradle.org/jcenter-shutdown.
> The Gradle Plugin Portal implicitly mirrors JCenter currently. If you’re using the Plugin Portal (via gradlePluginPortal() or the URL plugins.gradle.org/m2) to resolve your application’s dependencies, you may be relying on JCenter. You should avoid using the Plugin Portal as a repository, except for Gradle plugin projects.

A quick test on my part showed that fetching this dependency just redirects the request to jcenter (excess headers removed).
```curl
> curl -I -L https://plugins.gradle.org/m2/org/jetbrains/trove4j/trove4j/20160824/trove4j-20160824.pom
HTTP/1.1 303 See Other
Content-Length: 0
Connection: keep-alive
Location: https://jcenter.bintray.com/org/jetbrains/trove4j/trove4j/20160824/trove4j-20160824.pom

> curl -I -L https://plugins.gradle.org/m2/org/jetbrains/trove4j/trove4j
HTTP/1.1 303 See Other
Date: Wed, 24 Mar 2021 09:26:16 GMT
Content-Length: 0
Connection: keep-alive
Location: https://jcenter.bintray.com/org/jetbrains/trove4j/trove4j
```

pollend: umm, I see. I guess will either publish it ourselves or wait to see when jet-brains republishes the package :/

skaldarnar: See also https://github.com/MovingBlocks/TerasologyLauncher/pull/636#issuecomment-803668474

> We will be migrating the Plugin Portal away from JCenter before the final shutdown. Builds will not need to make changes while the Plugin Portal migrates away from JCenter.

skaldarnar (approved):

build.gradle
jdrueckert: FYI: On the topic of `trove4j` I found the following: https://youtrack.jetbrains.com/issue/IDEA-261387

PR #95, opened 2021-04-05
chore: upgrade gradle and rework build

pollend:

PR #96, opened 2021-04-15
chore: upgrade gradle to 6.8.2 from 5.4.1

keturn:
Gradle upgrade.

This does _not_ replace #95. That one targets a different branch.

This upgrade was mostly standard. The one thing that stood out is that the test modules had

```
jar {
    version = ''
}
```

and they've changed the name of the version property. I could have switched it to the new name, but it also defaults to `project.version` which seemed like it might be fine, so I took it out.

Is there anything that depends on the jar version being the empty string?


gestalt-module/src/main/java/org/terasology/gestalt/module/sandbox/JavaModuleClassLoader.java
```diff
@@ -154,8 +139,7 @@ private ClassLoader getBaseClassLoader() {
             if (permissionProvider.isPermitted(clazz)) {
                 return clazz;
             } else {
-                logger.error("Denied access to class (not allowed with this module's permissions): {}", name);
-                return null;
+                throw new ClassNotFoundException("Denied access to class (not allowed with this module's permissions): " + name);
```
keturn: FIXME: this should NOT have been in this PR

keturn:
> Is there anything that depends on the jar version being the empty string?

When it was the empty strings, the jars from the test modules get named like `moduleA.jar`.  
With the default, it is `moduleA-7.0.7-SNAPSHOT.jar`.

I haven't found anything that depends on the jar names being versionless, but I guess it clutters things a bit as versions change if the test-modules directory is not cleaned out.

DarkWeird:
> > Is there anything that depends on the jar version being the empty string?
> 
> When it was the empty strings, the jars from the test modules get named like `moduleA.jar`.
> With the default, it is `moduleA-7.0.7-SNAPSHOT.jar`.
> 
> I haven't found anything that depends on the jar names being versionless, but I guess it clutters things a bit as versions change if the test-modules directory is not cleaned out.

no one reference on versionless modules.
seems this was at gestalt origins.

DarkWeird:
Test `ApiScannerTest#test` failing

keturn:
> Test `ApiScannerTest#test` failing

Were they not before? Not sure what gradle version would have to do with that.

Failing tests addressed in #98

skaldarnar (approved):

PR #97, opened 2021-04-16
test: upgrade tests to JUnit 5

keturn:
Fixes #88 

- [x] asset-core
- [x] entity-system
- [x] es-perf
- [x] module
- [x] util

Depends on #96

PR #98, opened 2021-04-16
test: fix tests under java 11

keturn:
Fixes tests that fail under java 11 for ClassLoader.getURLs reasons.

Or fixes one test, anyway. I thought I saw other test failures earlier, but now running `test` in the gestalt project gets 321 passing and no skips or failures, so :shrug: 

Depends on #97


skaldarnar (approved):
Can confirm that `gradlew test` runs successful with these changes.

PR #101, opened 2021-04-20
feat: use Semver compatibility for dependency info

keturn:
Version 7 introduced the use of the Semver library (#80) for some functions of version handling. It changed the behavior of `getNextVersion` methods (as mentioned in its PR), and that left DependencyInfo returning an unsatisfiable version range when a module's `minVersion` targets a snapshot. (#100) https://github.com/MovingBlocks/gestalt/blob/a51ba072f2e7c8e084ddf0b530d73bbb713ae2a3/gestalt-module/src/main/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfo.java#L84-L89

Rather than try to re-implement the semantics of satisfying semantic versioning in the DependencyInfo and VersionRange classes, I propose using more of the Semver library's methods for this.

This PR is a step in that direction. It keeps the VersionRange class intact to preserve compatibility, but if a DependencyInfo was defined with _only_ a minimal version, instead of building a gestalt.VersionRange for it, we compile a Semver expression that will determine what matches.

**Dependencies:** #98. If that is not merged yet, you can review using this diff: https://github.com/MovingBlocks/gestalt/compare/v7/test/java11...v7/feat/semverComparison

## Questions for Review

- [x] Is the caret expression the right one?
- [ ] Is it worth keeping the existing VersionRange class at all, or should we convert entirely?
  - If we convert, be mindful of [VersionRangeTest](https://github.com/MovingBlocks/gestalt/blob/a51ba072f2e7c8e084ddf0b530d73bbb713ae2a3/gestalt-module/src/test/java/org/terasology/gestalt/naming/VersionRangeTest.java). lowerSnapshotInRange and higherSnapshotOutOfRange are the ones that are easy to miss. 
  - I only found _one_ place in Terasology that directly uses VersionRange, and it's easy to update.
- [ ] Instead of building an expression out of `minVersion`, should we allow module metadata to define its own expression?
  - We would need to use the Maven-style expressions in that case in order for compile-time dependency resolution to still work.

### DIFF
--- a/gestalt-module/src/main/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfo.java
+++ b/gestalt-module/src/main/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfo.java
@@ -1,26 +1,15 @@
-/*
- * Copyright 2019 MovingBlocks
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
 
 package org.terasology.gestalt.module.dependencyresolution;
 
 import org.terasology.gestalt.naming.Name;
+import org.terasology.gestalt.naming.SemverExpression;
 import org.terasology.gestalt.naming.Version;
 import org.terasology.gestalt.naming.VersionRange;
 
 import java.util.Objects;
+import java.util.function.Predicate;
 
 /**
  * Describes a dependency on a module. Dependencies apply to a range of versions - anything from the min version (inclusive) to the max version (exclusive) are supported.
@@ -124,6 +113,21 @@ public class DependencyInfo {
      */
     public VersionRange versionRange() {
         return new VersionRange(getMinVersion(), getMaxVersion());
+    }
+
+    public Predicate<Version> versionPredicate() {
+        if (maxVersion != null) {
+            return versionRange();
+        } else {
+            String source;
+            if (minVersion.isSnapshot()) {
+                // semver doesn't like snapshots in caret ranges?
+                source = minVersion + " | ^" + minVersion.getNextPatchVersion();
+            } else {
+                source = "^" + minVersion;
+            }
+            return new SemverExpression(source);
+        }
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


gestalt-module/src/main/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfo.java
```diff
@@ -126,6 +115,21 @@ public VersionRange versionRange() {
         return new VersionRange(getMinVersion(), getMaxVersion());
     }
 
+    public Predicate<Version> versionPredicate() {
```
keturn, line 118: TODO: doc

gestalt-module/src/main/java/org/terasology/gestalt/naming/SemverExpression.java
```diff
@@ -0,0 +1,32 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.gestalt.naming;
+
+import com.github.zafarkhaja.semver.expr.ExpressionParser;
+import com.google.common.base.MoreObjects;
+
+import java.util.function.Predicate;
+
+public class SemverExpression implements Predicate<Version> {
```
keturn, line 11: TODO: doc

gestalt-module/src/main/java/org/terasology/gestalt/naming/SemverExpression.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private final Predicate<com.github.zafarkhaja.semver.Version> expression;
+    private final String source;
+
+    public SemverExpression(String source) {
+        this.source = source;
+        this.expression = ExpressionParser.newInstance().parse(source);
```
keturn, line 18: TODO: error handling
or maybe using a documented exception here, and error handling in DependencyInfo?

keturn:
[gestalt snapshot dependencies](https://trello.com/c/h4hhzyKB/95-gestalt-snapshot-dependencies)

gestalt-module/src/main/java/org/terasology/gestalt/module/dependencyresolution/ResolutionAttempt.java
```diff
@@ -125,7 +112,7 @@ private void populateConstraints() {
                         Module versionedModule = registry.getModule(name, version.getVersion().get());
                         DependencyInfo info = versionedModule.getMetadata().getDependencyInfo(dependency);
                         if (info != null) {
-                            constraintTable.put(version.getVersion().get(), new CompatibleVersions(info.versionRange(), info.isOptional() && !optionalStrategy.isRequired()));
+                            constraintTable.put(version.getVersion().get(), new CompatibleVersions(info.versionPredicate(), info.isOptional() && !optionalStrategy.isRequired()));
```
skaldarnar, line 115: I'd love to have a uniform code formatter...

skaldarnar (approved):

skaldarnar:
I leave the two open questions for later. I do believe that following any _common_ way to define versions would be a reasonable choice (be it similar to [npm's advanced range syntax](https://docs.npmjs.com/cli/v7/using-npm/semver#advanced-range-syntax) or resembling [gradle's version declarations](https://docs.gradle.org/current/userguide/single_versions.html)).

gestalt-module/src/main/java/org/terasology/gestalt/module/dependencyresolution/ResolutionAttempt.java
keturn: what prompted that thought on this line in particular?

skaldarnar: hm, probably the length of that line :thinking:

PR #102, opened 2021-04-24
fix(AutoReloadAssetTypeManager): synchronize methods to avoid NPE

keturn:
switchEnvironment leaves a closed reloadOnChangeHandler for a moment, which caused reloadChangedAssets() to crash if called from another thread.

I don't believe this is the _most efficient_ locking strategy for this, but the ergonomics of marking methods as `synchronized` are _so much better_ than using `Lock.lock()`. I tried a few other things but they added a ton of noise to the code, so I decided to submit this more readable version until we find something wrong with it.

I have very little experience managing thread safety in Java and am extra open to suggestions on this one.

Blocks https://github.com/MovingBlocks/Terasology/pull/4622

## Test plan

MTE tests run under https://github.com/MovingBlocks/Terasology/pull/4622 consistently crash without a fix for this. Should be able to reproduce with something like

```sh
./gradlew :modules:Health:integrationTest
```

### DIFF
--- a/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/module/autoreload/AutoReloadAssetTypeManager.java
+++ b/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/module/autoreload/AutoReloadAssetTypeManager.java
@@ -1,23 +1,9 @@
-/*
- * Copyright 2019 MovingBlocks
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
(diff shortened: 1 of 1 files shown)


jdrueckert (approved):
Sounds reasonable to me. However, I likely have even less experience with thread safety in Java, so feel free to wait for a review from somebody else :sweat_smile:

keturn:
> Sounds reasonable to me. However, I likely have even less experience with thread safety in Java, so feel free to wait for a review from somebody else 😅

It's a topic that's worth some attention, but I get the feeling I should not hold my breath for it.

I rate this as a low-risk change, so I'm going to go ahead and merge it.

If we find `reloadChangedAssets` causing things to stutter we can come back and give it another look.

PR #103, opened 2021-04-25
ci: add Jenkinsfile

keturn:
Based on the release/v5.x branch with a few tweaks.

The build and test stages run okay. The publish stage doesn't publish PR branches, which is good. I guess we'll see if it publishes a snapshot once it is merged?

Differences from #95:
- this targets v7, not develop
- this includes _only_ Jenkinsfile, not other build changes
- the Jenkinsfile has some things ordered differently and runs `gradlew --parallel` in some places, but it is overall compatible. It runs the same gradle tasks and uses the same conditions for when to publish something.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,19 +6,27 @@ pipeline {
         stage('Build') {
             steps {
                 sh './gradlew --info --console=plain --parallel assemble compileTest'
-                recordIssues enabledForFailure: true, tools: [java()]
+            }
+            post {
+                always {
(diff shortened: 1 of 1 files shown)


keturn:
This was a bit rocky when I had the tests running with `gradlew --parallel`, but it seems to be behaving okay with that step run in serial.

Cervator:
[Do next Gestalt release / snapshot with new Gradle / Jenkins stuff](https://trello.com/c/CrJ9oZQM/89-do-next-gestalt-release-snapshot-with-new-gradle-jenkins-stuff)

PR #104, opened 2021-04-25
ci: build with java 11? [spike]

keturn:
Experimenting to see how we could keep building with whatever lowest-common-denominator JDK we need, but also run the test suite under a newer JDK.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,6 @@
 pipeline {
     agent {
-        label "java8"
+        label "java11"
     }
     stages {
         stage('Build') {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,15 +1,28 @@
 pipeline {
-    agent {
-        label "java11"
-    }
+    agent none
+    stages {
+        stage("Validate") {
+            matrix {
+                axes {
+                    axis {
+                        name 'JAVA_VERSION'
+                        values 'java8', 'neo-java'
+                    }
+                }
+                agent {
+                    label env.JAVA_VERSION
+                }
     stages {
         stage('Build') {
             steps {
+                sh './gradlew tasks # make sure gradle-wrapper has installed'
+                sh "( ./gradlew --version ; ./gradlew projects buildEnvironment ) | tee build-environment-${JAVA_VERSION}.log"
                 sh './gradlew --info --console=plain --parallel assemble compileTest'
             }
             post {
                 always {
-                    recordIssues enabledForFailure: true, tools: [java()]
+                    recordIssues enabledForFailure: true, tools: [java(id:"java-$JAVA_VERSION", name:"Compilation $JAVA_VERSION")]
+                    archiveArtifacts artifacts: "build-environment-${JAVA_VERSION}.log"
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -1,15 +1,28 @@
...
+        stage("Validate") {
+            matrix {
+                axes {
+                    axis {
+                        name 'JAVA_VERSION'
+                        values 'java8', 'neo-java'
+                    }
+                }
+                agent {
+                    label env.JAVA_VERSION
```
keturn, line 13: I ended up using the labels as the axis values, but that doesn't feel great.

How would I do a map from values to labels in a declarative pipeline?

Or maybe this _is_ the way to go, and it's just a matter of naming labels consistently?

Jenkinsfile
```diff
@@ -23,8 +36,8 @@ pipeline {
                 always {
                     junit testResults: '**/build/test-results/test/*.xml'
                     recordIssues tools: [
-                      javaDoc(),
-                      taskScanner(includePattern: '**/*.java,**/*.groovy,**/*.gradle,**/*.kts', lowTags: 'WIBNIF', normalTags: 'TODO, FIXME', highTags: 'ASAP')
+                      javaDoc(id:"javaDoc-$JAVA_VERSION", name: "JavaDoc $JAVA_VERSION"),
+                      taskScanner(id:"tasks-$JAVA_VERSION", name: "Tasks $JAVA_VERSION", includePattern: '**/*.java,**/*.groovy,**/*.gradle,**/*.kts', lowTags: 'WIBNIF', normalTags: 'TODO, FIXME', highTags: 'ASAP')
```
keturn, line 40: taskScanner works purely on the source text so it's not really useful to include in the matrix. Doing it once would be sufficient.

Jenkinsfile
```diff
@@ -33,6 +46,7 @@ pipeline {
         }
         stage('Publish') {
             when {
+                expression { env.JAVA_VERSION == 'java8' }
```
keturn, line 49: I put this in the matrix so it can keep going on the java8 node it used to build.

(Not publishing a java11 build because that sounds like a different kettle of worms.)

But maybe this should be outside the matrix so it doesn't happen unless the java11 tests succeed?

Cervator:
[Separate Jenkins test stages](https://trello.com/c/fgorgiQY/53-separate-jenkins-test-stages)

Cervator:
This is interesting, nice work 👍 

I haven't tried using matrix in pipeline land at all, be it declarative or scripted. Additionally I've started using _multiple_ labels as requirements, and I'm not sure how those two approaches would get along. Not that we can't just go to single (perhaps more detailed) labels. I need more time to dig into this some to really be able to contribute much of anything sensible I think, and there are a few things in front of it :-)

keturn:
>  Additionally I've started using multiple labels as requirements, and I'm not sure how those two approaches would get along.

Using `label "lib && light && $JAVA_VERSION"` seems to work. At least for `java8` where it can actually find something that fits. Presumably we could make a `java11` to fit as well.

Cervator:
This deserves more time from me but I dunno when that'll be. I only remember matrix from _way_ long ago, where it had some warts. Without really having the knowledge to judge its state today - I do wonder if it would be the best approach, especially if there are some stages where it doesn't make sense to spread across the full matrix. The main alternative I can think of is just clever use of stages and parallel but with hard coded targets (dunno if scripted would have more flexibility there or not needed). Not sure if we're likely to go much past feeding 2 JDKs to the matrix?

Cervator:
[More ci pipeline improvements](https://trello.com/c/pg3mdled/116-more-ci-pipeline-improvements)

soloturn:
this works nowadays.

PR #105, opened 2021-04-25
Release v5.1.6 and minor logistics tweaks

Cervator:
Other than the minor non-functional changes in this PR I _believe_ this would also finally be a release for https://github.com/MovingBlocks/gestalt/commit/f92e9cfd2bb257330eb6bdaf7d82d84d1d34a963 by @keturn from some time ago. TS engine hasn't been able to use that yet but if still relevant (maybe not with v7 support on the horizon) then with this released it should be able to.

Tested both locally and via http://jenkins.terasology.io/teraorg/job/Nanoware/job/Gestalt/job/release%252Fv5.x/ - although "tested" insofar as it builds, haven't tried to actually _use_ it with anything.

### DIFF
--- a/gradle/common.gradle
+++ b/gradle/common.gradle
@@ -54,11 +54,6 @@ check.dependsOn rootProject.extractAnalyticsConfig
 
 // We use both Maven Central and our own Artifactory instance, which contains module builds, extra libs, and so on
 repositories {
-    // For development so you can publish binaries locally and have them grabbed from there
-    mavenLocal()
-
-    // External libs - jcenter is Bintray and is supposed to be a superset of Maven Central, but do both just in case
-    jcenter()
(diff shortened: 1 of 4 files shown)


gradle/common.gradle
```diff
@@ -54,11 +54,6 @@ check.dependsOn rootProject.extractAnalyticsConfig
 
 // We use both Maven Central and our own Artifactory instance, which contains module builds, extra libs, and so on
 repositories {
-    // For development so you can publish binaries locally and have them grabbed from there
-    mavenLocal()
```
keturn, line 58: bye mavenLocal!

keturn (approved):
I didn't really expect another 5.x release, but sure, why not!

PR #109, opened 2021-04-26
chore: update asset scheme

pollend:
this changes how assets work with the assetmanager and attempts to address some problems.

1. for transient assets the current scheme will slowly leak weak pointers since they are not properly removed from the multimap. 
2. there is now just a single map from a resource urn to the start of a chain of assets. (non-transient assets use soft-reference and transient assets use a weakreference but weak references use a strong reference so the non-transient asset so it's not disposed unless all transient assets have been disposed in the chain.  
  - the non-transient asset is always placed first and all transient assets use a weak reference
  - each asset is chained from the beginning forward as new assets are added they are placed at the beginning of the chain unless if its a non-transient asset then it skips the first entry
  
image

### DIFF
--- a/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
+++ b/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
@@ -102,7 +102,7 @@ public abstract class Asset<T extends AssetData> {
         if(rootNode == null) {
             return Collections::emptyIterator;
         }
-        return () -> new Iterator<>() {
+        return () -> new Iterator<WeakReference<Asset<T>>>() {
             AssetNode<T> node = null;
 
             @Override
--- a/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
+++ b/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
@@ -70,20 +70,6 @@ public abstract class Asset<T extends AssetData> {
         SoftReference<Asset<T>> reference;
         AssetNode<T> next;
 
-        protected void clearParent() {
-            Asset<T> instance = reference == null ? null: reference.get();
-            if(instance != null) {
-                instance.parent = null;
-            }
-        }
-
-        protected void setParent(Asset<T> parent) {
-            Asset<T> instance = reference == null ? null: reference.get();
-            if(instance != null) {
-                instance.parent = parent;
-            }
-        }
-
         protected boolean hasValidAsset() {
             Asset<T> instance = reference == null ? null: reference.get();
             if(instance == null) {
--- a/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
+++ b/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
@@ -1,19 +1,5 @@
-/*
- * Copyright 2019 MovingBlocks
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
 package org.terasology.gestalt.assets;
 
 import com.google.common.base.Preconditions;
@@ -77,16 +63,20 @@ public abstract class Asset<T extends AssetData> {
         }
     }
 
-    protected Iterable<WeakReference<Asset<T>>> instances() {
+    /**
+     * return all the instances that are associated with this type of asset.
+     * @return instances
+     */
+    protected final Iterable<WeakReference<Asset<T>>> instances() {
         if (this.parent != null) {
             return parent.instances();
         }
         return instances;
     }
 
-    protected void cleanup() {
-        if(this.parent != null) {
-            parent.cleanup();
+    private void compactInstances() {
+        if (this.parent != null) {
+            parent.compactInstances();
             return;
         }
         Iterator<WeakReference<Asset<T>>> iter = instances.iterator();
@@ -182,6 +172,11 @@ public abstract class Asset<T extends AssetData> {
         return (Optional<U>) assetCopy;
     }
 
+    /**
+     * return the non-instanced version of this asset. if the asset is already the normal
+     * type then it returns itself.
+     * @return non-instanced version of this Asset
+     */
     public final Asset<T> getNormalAsset() {
         if(parent == null) {
             return this;
@@ -194,6 +189,7 @@ public abstract class Asset<T extends AssetData> {
      */
     public final synchronized void dispose() {
         if (!disposed) {
+            compactInstances();
             disposed = true;
             assetType.onAssetDisposed(this);
             disposalHook.dispose();
--- a/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
+++ b/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
@@ -54,15 +54,15 @@ import java.util.Optional;
 @ThreadSafe
 public abstract class Asset<T extends AssetData> {
 
+    AssetNode<Asset<T>> next;
+    Asset<?> parent;
+
     private final ResourceUrn urn;
     private final AssetType<?, T> assetType;
     private final DisposalHook disposalHook = new DisposalHook();
     private volatile boolean disposed;
 
-    protected AssetNode<Asset<T>> next;
-    protected Asset<?> parent;
-
-    protected static class AssetNode<U extends Asset<?>> {
+    static class AssetNode<U extends Asset<?>> {
         public AssetNode(U root,U instance) {
             this.reference = new SoftReference<>(instance);
             instance.parent = root;
--- a/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
+++ b/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
@@ -175,10 +175,11 @@ public abstract class Asset<T extends AssetData> {
 
     /**
      * return the non-instanced version of this asset. if the asset is already the normal
-     * type then it returns itself.
+     * type then it returns itself. instanced assets are temporary copies from the normal loaded instances.
+     *
      * @return non-instanced version of this Asset
      */
-    public final Asset<T> getConcreteAsset() {
+    public final Asset<T> getNormalAsset() {
         if (parent == null) {
             return this;
         }
@@ -192,7 +193,6 @@ public abstract class Asset<T extends AssetData> {
         if (!disposed) {
             compactInstances();
             disposed = true;
-            assetType.onAssetDisposed(this);
             disposalHook.dispose();
             if(parent == null) {
                 for (WeakReference<Asset<T>> inst : this.instances()) {
--- a/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
+++ b/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
@@ -17,14 +17,13 @@
 package org.terasology.gestalt.assets;
 
 import com.google.common.base.Preconditions;
-
 import net.jcip.annotations.ThreadSafe;
-
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.terasology.context.annotation.API;
 
-import java.lang.ref.PhantomReference;
-import java.lang.ref.ReferenceQueue;
 import java.lang.ref.SoftReference;
+import java.security.PrivilegedActionException;
 import java.util.Optional;
 
 /**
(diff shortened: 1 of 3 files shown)


gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
```diff
@@ -56,6 +59,40 @@
     private final DisposalHook disposalHook = new DisposalHook();
     private volatile boolean disposed;
 
+    protected AssetNode<Asset<T>> next;
```
immortius: Why are these protected? Are they intended to be messed with by subtypes? For what purpose?

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -642,13 +717,14 @@ public String toString() {
         }
     }
 
-    private static final class AssetReference<T> extends PhantomReference<T> {
+    public final class AssetReference<T extends Asset<?>> extends PhantomReference<T> {
```
immortius: Why is this public now? Why is it no longer static? There's no reason other classes should be doing anything with an asset reference. And AssetReference is not doing anything with it's parent asset type?

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -581,7 +656,7 @@ public boolean isLoaded(ResourceUrn urn) {
      * @return A list of all the loaded assets.
      */
     public Set<T> getLoadedAssets() {
-        return ImmutableSet.copyOf(loadedAssets.values());
+        return ImmutableSet.copyOf(loadedAssets.values().stream().map(Reference::get).filter(Objects::nonNull).collect(Collectors.toSet()));
```
immortius: There's no need to make an ImmutableSet copy and them stream the values into a new set could just stream into the set.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -75,8 +75,7 @@
     private final Class<U> assetDataClass;
     private final AssetFactory<T, U> factory;
     private final List<AssetDataProducer<U>> producers = Lists.newCopyOnWriteArrayList();
-    private final Map<ResourceUrn, T> loadedAssets = new MapMaker().concurrencyLevel(4).makeMap();
-    private final ListMultimap<ResourceUrn, WeakReference<T>> instanceAssets = Multimaps.synchronizedListMultimap(ArrayListMultimap.<ResourceUrn, WeakReference<T>>create());
```
immortius, line 79: I don't understand why the fix for the leakage isn't just iterating through the values of the resource urn and cleaning up the null references on disposal rather than this... complex manual linked-list implementation.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -153,23 +157,23 @@ public synchronized boolean isClosed() {
...
      */
     public synchronized void disposeAll() {
-        loadedAssets.values().forEach(T::dispose);
-
-        for (WeakReference<T> assetRef : ImmutableList.copyOf(instanceAssets.values())) {
-            T asset = assetRef.get();
-            if (asset != null) {
+        loadedAssets.values().forEach(k -> {
+            Asset<U> asset = k.get();
+            if(asset != null) {
```
immortius: Need a space between the if and (
:P

immortius: It probably isn't impossible for the parent asset to be disposed while instance assets exist - this check may leave those undisposed.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
immortius: Package-private might be more correct.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -267,33 +274,95 @@ public synchronized void clearProducers() {
...
+                targetAsset.next = new Asset.AssetNode<>(targetAsset, current);
+                Asset.AssetNode<Asset<U>> next = targetAsset.next;
+                while (next != null) {
+                    next.setParent(targetAsset); // update all parents to point to new target
+                    next = next.next;
+                }
+                loadedAssets.put(parentUrn, targetUrn.isInstance() ? new WeakReference<T>(targetAsset) : new SoftReference<T>(targetAsset));
+            }
         } else {
-            loadedAssets.remove(asset.getUrn());
```
immortius, line 278: Removing this statement means loaded assets will contain disposed assets... which means they won't be reloaded automatically if someone requests it again by Urn.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -267,33 +274,95 @@ public synchronized void clearProducers() {
...
 
+
     /**
-     * Notifies the asset type when an asset is created
+     * Notifies the asset type when an asset is disposed
      *
-     * @param asset The asset that was created
+     * @param asset The asset that was disposed.
      */
-    synchronized void registerAsset(Asset<U> asset, DisposalHook disposer) {
```
immortius: The github diff is really confused and I cannot tell if you've made changes to registerAsset or not.

immortius (changes requested):
Overall, this request needs work. I'm concerned firstly with the scope of the changes. I think it is a bit overdone - in particular I do not believe a node based linked list system needed to be developed, as simpler tweaks around the existing implementation would have enabled the improvements. Also I'm concerned some changes introduce bugs or break encapsulation. Also some style issues. 

I do agree with:
- Changing normal assets to be kept via soft references
- Cleaning up null weak references from the instance asset map
- Adding tracking of parent urn to the asset references to enable the above

pollend:
> Overall, this request needs work. I'm concerned firstly with the scope of the changes. I think it is a bit overdone - in particular I do not believe a node based linked list system needed to be developed, as simpler tweaks around the existing implementation would have enabled the improvements. Also I'm concerned some changes introduce bugs or break encapsulation. Also some style issues.
> 
> I do agree with:
> 
>     * Changing normal assets to be kept via soft references
> 
>     * Cleaning up null weak references from the instance asset map
> 
>     * Adding tracking of parent urn to the asset references to enable the above 

One of the benefits with this scheme is it decouples a lot of the book keeping from AssetType. Assets can clone and insert a reference into the chain without depending on or notifying assetType of the change. at most you would just need to introduce a way to lock the urn which isn't too hard to add to the current code. an AssetType would just manage the bookkeeping and clean up references as they go stale. The implementation could be cleaned up quite a bit since there is a large amount of book keeping with the code I added. The PR is pretty rough since the implementation is quite a bit more complicated.

pollend:
@immortius so I think this is a lot simpler then the original system. I decoupled the ownership of instanced assets from assetTypes. so assetTypes now only track non instanced assets. assets that create instances will now push it to a linked list. reloading just follows the chain of instanced assets and since each instance also has a strong link back to the parent asset then that resource is only held on as long as instanced assets are still valid. if a parent becomes invalid then all instanced assets should become invalid as well :D.

pollend:
[Update Asset Scheme for gestalt](https://trello.com/c/k6xQtkA8/113-update-asset-scheme-for-gestalt)

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
```diff
@@ -150,6 +108,7 @@ protected void cleanup() {
     public Asset(ResourceUrn urn, AssetType<?, T> assetType) {
```
immortius: Just noting some cleanup for style is required - constructors should be the top methods in a class, I see a couple of places where there are spaces missing between if or while and the bracket.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
```diff
@@ -121,12 +166,30 @@ public final synchronized void reload(T data) {
...
+            }
+        }
+        assetCopy.ifPresent(this::appendInstance);
+        return (Optional<U>) assetCopy;
     }
 
-    final synchronized Optional<? extends Asset<T>> createCopy(ResourceUrn copyUrn) {
-        Preconditions.checkState(!disposed);
-        return doCreateCopy(copyUrn, assetType);
+    public final Asset<T> getNormalAsset() {
```
immortius, line 182: Please add a javadoc to this method (all public methods should have javadoc to assist a user.
Not clear from the name what this does - what makes an asset "normal"? I might suggest getArchetype() or something like that, not sure if that adds clarity though.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
immortius: This is still a problem, although the solution is probably a little different now. Need to be able to reload a disposed asset if it is requested again (so as not to lose track of instances)

immortius: Or maybe we just dispose all instances with the parent.

immortius (changes requested):

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
```diff
@@ -50,12 +56,49 @@
...
+    }
+
+    protected Iterable<WeakReference<Asset<T>>> instances() {
+        if (this.parent != null) {
+            return parent.instances();
+        }
+        return instances;
+    }
+
+    protected void cleanup() {
```
immortius: Can this be private? Since resource cleanup should happen in the disposal hook, what reason is there for implementors to change this?

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
```diff
@@ -50,12 +56,49 @@
...
+            Preconditions.checkArgument(!getUrn().isInstance());
+            asset.parent = this;
+            instances.add(new WeakReference<>(asset));
+        } else {
+            asset.parent = parent;
+            parent.instances.add(new WeakReference<>(asset));
+        }
+    }
+
+    protected Iterable<WeakReference<Asset<T>>> instances() {
```
immortius: Can this be private/what reason is there for implementors to override this?

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
immortius: Or call it.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
immortius: Or call it.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -288,15 +269,45 @@ synchronized void registerAsset(Asset<U> asset, DisposalHook disposer) {
         if (closed) {
             throw new IllegalStateException("Cannot create asset for disposed asset type: " + assetClass);
         } else {
-            if (asset.getUrn().isInstance()) {
-                instanceAssets.put(asset.getUrn(), new WeakReference<>(assetClass.cast(asset)));
-            } else {
-                loadedAssets.put(asset.getUrn(), assetClass.cast(asset));
+            if(!asset.getUrn().isInstance()) {
```
immortius: This is just a general style comment - should be a space after if, happens in a number of changes in this PR. Needs a style cleanup.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -288,15 +269,45 @@ synchronized void registerAsset(Asset<U> asset, DisposalHook disposer) {
...
+    private void compactResource(ResourceUrn target) {
+        Preconditions.checkArgument(!target.isInstance());
+        if(!loadedAssets.containsKey(target)) {
+            return;
+        }
+
+        Reference<T> reference = loadedAssets.get(target);
+        Asset<U> current = reference.get();
+        if(current != null && current.isDisposed() && Iterables.isEmpty(current.instances())) {
+            logger.warn("non instanced asset is disposed with instances. instances will become orphaned.");
```
immortius: Probably would be cleaner to dispose the children when the parent is disposed. But maybe for a future PR.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -288,15 +269,45 @@ synchronized void registerAsset(Asset<U> asset, DisposalHook disposer) {
...
+        }
+
+        if(current == null || current.isDisposed()) {
+            // disposing of a non instanced asset will orphan the instanced assets.
+            loadedAssets.remove(target);
+            return;
+        }
+        current.cleanup();
```
immortius: This line probably is unreachable without an asset being loaded fresh into the urn in the meantime. Which I guess is possible through the onAssetDisposed pathway, since it isn't synchronized. Not sure it is a meaningful action regardless.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -640,15 +646,17 @@ public boolean unlock() {
         public String toString() {
             return "lock(" + urn + ")";
         }
+
     }
 
-    private static final class AssetReference<T> extends PhantomReference<T> {
+    private final class AssetReference<T extends Asset<?>> extends PhantomReference<T> {
```
immortius: Why is this not static? The implicit reference to the outer object doesn't seem to be used.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -266,37 +245,60 @@ public synchronized void clearProducers() {
...
      */
     void onAssetDisposed(Asset<U> asset) {
-        if (asset.getUrn().isInstance()) {
-            instanceAssets.get(asset.getUrn()).remove(new WeakReference<>(assetClass.cast(asset)));
-        } else {
-            loadedAssets.remove(asset.getUrn());
+        if (!asset.getUrn().isInstance()) {
+            disposeAsset(asset.getUrn());
         }
     }
```
pollend: @immortius I don't think we need this onAssetDisposed call anymore.  looks like the asset is already managed in the processDisposal call.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
jdrueckert: @pollend I think it's still a bit unclear what "the normal type" here is 🤔

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
jdrueckert: @pollend I think this was not yet resolved, was it?

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
jdrueckert: @pollend I think this question was not answered, was it? I see that you made it `final`, but why does it need to be on `protected` level?

jdrueckert: @pollend this question is still open

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
jdrueckert: `getConcreteAsset` doesn't really make this clearer to me.
please add a comment in code that explains what characterizes an asset as "concrete"

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
pollend: you should be able to access all the list of instances from within the asset.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
pollend: I reverted it back to normal asset and updated the javadocs.

PR #111, opened 2021-04-27
test(gestalt-module): tests for DependencyInfo.versionPredicate

keturn:
This was intended to be included with #101 but it fell off the truck somewhere.

### DIFF
--- a/gestalt-module/src/test/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfoTest.java
+++ b/gestalt-module/src/test/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfoTest.java
@@ -1,0 +1,74 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.gestalt.module.dependencyresolution;
+
+import org.junit.jupiter.api.Test;
+import org.terasology.gestalt.naming.Version;
+
+import java.util.function.Predicate;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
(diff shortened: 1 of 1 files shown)


gestalt-module/src/test/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfoTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+        Version nextVersion = minVersion.getNextPatchVersion();
+        assertTrue(predicate.test(nextVersion), "failed to match the next patch version");
+        assertTrue(predicate.test(new Version(nextVersion + "-SNAPSHOT")), "failed to match the next snapshot");
+
+        Version nextMinorVersion = minVersion.getNextMinorVersion();
+        assertTrue(predicate.test(nextMinorVersion), "failed to match the next minor version");
+        assertTrue(predicate.test(new Version(nextMinorVersion + "-SNAPSHOT")), "failed to match the next minor snapshot");
+
+        assertFalse(predicate.test(new Version("1.4.6")), "inappropriately matched an earlier patch version");
```
skaldarnar, line 53: Should we add a test to ensure that a snapshot of a non-snapshot version is not matched? Probably same for the test case above. 

```suggestion
        assertFalse(predicate.test(new Version("1.4.7-SNAPSHOT")), "inappropriately matched same version snapshot");
```

skaldarnar (approved):

gestalt-module/src/test/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfoTest.java
jdrueckert: will be addressed in PR for #113

Issue #15, opened 2015-04-20
Give feedback on failed dependency resolution


msteiger:
Currently, the `ResolutionResult` class returns `false` when the resolutions fails. I would find it helpful to get more information on why that happened.

This could include the modules (version ranges) in the DAG that were missing or the dependency links (as modeled by the `Constraint` class).

Cervator:
This could also be what we need to give the user more choice in the game setup menu - accepting newer module dependency versions after a disclaimer, for instance. Sounded like @Josharias was interested in helping out with that in the game menu piece :-)

immortius:
Hmm. To be honest I had thrown this one in the too hard basket to worry
about much later. I'm sure something is possible, although I think it might
be separate from the Dependency Resolver.

The problem with the algorithm used is it isn't clear to me how to detect
the root cause of a resolution failure - and in fact there may be multiple
root causes. Constraints don't fail - the are merely statements used to
filter out options that can't be satisfied - and any given filtering out
may be fine. It isn't until every version of a desired module is eliminated
that you can be sure there is an issue, but then the cause isn't the
constraint that eliminates the last version, but the initial constraints
triggering the chain leading to the elimination of each version.

keturn:
> To be honest I had thrown this one in the too hard basket to worry about much later.

_checks watch_ ⌚

Is five years later “much later” yet? :wink: 

but seriously, this is a _huge_ source of frustration in trying to work in Terasology. You just get this message that says “failed” with nothing to follow up on and your only options are to walk away or try slamming your face into the wall again and hoping it doesn't hurt this time.

So after spending a couple hours with a debugger, I'm here in ResolutionAttempt.​VersionConstraint.​ConstrainFrom, figuring this is where stuff gets thrown out. (which is itself information! it was loaded in the first place and then excluded by a constraint, which is different than never-found.)

and I've got the debugger spitting this out when something's invalid

```java
String.format("INCOMPATIBLE %s %s %s %s %s", 
    fromVersions, this, toVersions, 
    version.version.orElse(null), compatibility.versionRange);
```

which is starting to look _almost_ useful, except I can't read this very well

```[org.terasology.module.ResolutionAttempt$PossibleVersion@e9495]```

so I think the bare minimum here is:

* implement PossibleVersion.toString
* implement CompatibleVersions.toString
* log something here at DEBUG severity when a constraint changes something
  * probably with an extra flag if that removal left only the empty set behind

[...]

PR #115, opened 2021-05-04
feat!: make version comparison semver compliant

### LINKED ISSUE #113: change VersionRange comparison so 2.0-SNAPSHOT does not satisfy 2.0
We've decided to change version comparison to be more consistent with Semantic Versioning.

Currently a dependency declared as `2.0.0` may be satisfied by a module implementing `2.0.0-SNAPSHOT`. This should not be true under semantic versioning.

Requires a change to the VersionRange class and tests, as well as documentation. (Check both Terasology and gestalt wikis.)

jdrueckert:
Closes #113


skaldarnar:
- [ ] I do have the Terasology Wiki changes lined up locally and can push anytime this is merged.
- [x] https://github.com/MovingBlocks/gestalt/wiki/Dependency-Resolution does not have any specific information about how versions are resolved.
- [ ] @jdrueckert we should probably improve the JavaDoc for `VersionRange` and `DependencyInfo`.

jdrueckert:
[[GESTALTv7] migrate module & asset](https://trello.com/c/te4uIhaz/46-gestaltv7-migrate-module-asset)

gestalt-module/src/main/java/org/terasology/gestalt/naming/VersionRange.java
```diff
@@ -44,7 +44,7 @@ public Version getUpperBound() {
      * @return Whether version falls within the range
      */
     public boolean contains(Version version) {
-        return version.compareTo(lowerBound.getSnapshot()) >= 0 && version.compareTo(upperBound.getSnapshot()) < 0;
+        return version.compareTo(lowerBound) >= 0 && version.compareTo(upperBound.getSnapshot()) < 0;
```
keturn, line 47: This works for Terasology's current use cases. I agree that having one VersionRange class is better than having the one-foot-in-one-foot-out thing I had going on.

The edge cases this falls apart on are things like `1.4.7-SNA` or `1.4.7-SNAPSHOT-2`. Which—maybe we don't care? Would be fine with declaring that out of scope. But that is the sort of thing that motivated me to try to delegate the problem to the SemVer library (since we have it anyway).

gestalt-module/src/main/java/org/terasology/gestalt/naming/Version.java
```diff
@@ -108,6 +108,13 @@ public Version getNextPatchVersion() {
         return new Version(semver.incrementPatchVersion());
     }
 
+    /**
+     * Returns the version core according to semver, i.e. the purely numerical version without any pre-release or build metadata info
+     */
+    public Version getVersionCore() {
```
keturn: thinking about naming for this one. Don't blame you for keeping JSemVer's term for this as "normal," as that adjective doesn't communicate much.

Maybe "base version"?

just brainstorming. "Core" is okay too.

gestalt-module/src/main/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfo.java
```diff
@@ -109,25 +108,12 @@ public void setOptional(boolean optional) {
     }
 
     /**
-     * @return The range of supported versions
+     * Returns a predicate that yields true when applied to version that is within the version range described by this dependency information
+     *
+     * @return
```
keturn: nit: empty tag

gestalt-module/src/main/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfo.java
```diff
@@ -73,9 +72,9 @@ public void setMinVersion(Version value) {
     public Version getMaxVersion() {
         if (maxVersion == null) {
             if (minVersion.getMajor() == 0) {
-                return minVersion.getNextMinorVersion();
+                return minVersion.getVersionCore().getNextMinorVersion();
             }
-            return minVersion.getNextMajorVersion();
+            return minVersion.getVersionCore().getNextMajorVersion();
```
keturn: thought: this is another one of those places that makes me think "if we have to keep this `if "0.x"` logic ourselves, what good is having JSemVer?"

keturn: suggestion (non-blocking): regardless of JSemVer, we could move this to a method on Version. "getExclusiveUpperCompatibleRangeVersion" or somesuch.

or maybe instead of trying to figure out what to name the method that returns the exclusive-upper-bound-version, have Version construct the VersionRange itself.
```java
public VersionRange getCompatibleRange()
```

keturn (changes requested):
This abandons the attempt at moving the compatible-version logic to the JSemVer library. But since writing that I have found out that gradle's default version-compatibility logic is not nearly as semver-compatible as I thought it was. VersionRange is a pretty good model as a thing to translate to gradle after all, as gradle doesn't have the `^` range comparison operator.

I marked a nit or two but I am good with the overall approach.

Suggestion: I think this got rid of the things that needed Version.semver to be package-visible: https://github.com/MovingBlocks/gestalt/blob/511e6b22ade650fd8f09b20f1f56497f88d01246/gestalt-module/src/main/java/org/terasology/gestalt/naming/Version.java#L22  
so we can make that private again and revert 4b6ea6d50176da0ac0d93a9186f1b44ba2cb12c4  (should do before release)

gestalt-module/src/main/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfo.java
```diff
@@ -109,25 +108,12 @@ public void setOptional(boolean optional) {
...
-     * @return The range of supported versions
+     * Returns a predicate that yields true when applied to version that is within the version range described by this dependency information
+     *
+     * @return
      */
-    public VersionRange versionRange() {
-        return new VersionRange(getMinVersion(), getMaxVersion());
-    }
-
     public Predicate<Version> versionPredicate() {
```
keturn, line 113: Thinking about types here.

`Predicate<Version>` is sufficient for gestalt ResolutionAttempt to do its thing. But there are a few other requirements we should consider:

* for troubleshooting purposes, this should at _least_ have a `toString` value that makes more sense than `lambda@123` or whatever. (#15 might also appreciate a rich type, but will probably find a string sufficient.)
* for gradle compatibility, it needs _either_ this minVersion, maxVersion pair **or** a string containing a maven-compatible range descriptor (which is what VersionRange.toString currently provides)

which makes me think we want to keep the return type as VersionRange after all.

gestalt-module/src/main/java/org/terasology/gestalt/naming/VersionRange.java
skaldarnar: That sounds like completely removing `VersionRange` would be another option.

Technically, SemVer does not describe anything special about `-SNAPSHOT` (I think this is a maven concept). All of the `-....` exmaples are pre-releases, and none of them matches a version requirement for the released version.

gestalt-module/src/main/java/org/terasology/gestalt/naming/Version.java
skaldarnar: This was derived from the semver specification directly:

```
<valid semver> ::= <version core>
                 | <version core> "-" <pre-release>
                 | <version core> "+" <build>
                 | <version core> "-" <pre-release> "+" <build>
```

Maybe `getCoreVersion` feels a bit more natural, but is still close enough to the name in the spec?

gestalt-module/src/main/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfo.java
skaldarnar: thanks auto-formatter 🧐

gestalt-module/src/main/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfo.java
skaldarnar: 🤔 noted - have to think about this.

gestalt-module/src/main/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfo.java
skaldarnar: Let me bring up my question about sum types again... 🙄  nargh

keturn: I don't think this one needs sum types. One class can implement both `Predicate<Version>` and `ConvertableToMavenRange` or whatever.

gestalt-module/src/main/java/org/terasology/gestalt/naming/Version.java
skaldarnar: renamed to `getCoreVersion` in f41d2f1

gestalt-module/src/main/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfo.java
skaldarnar: removed in f41d2f1

skaldarnar:
@keturn are any of the remaining comments blocking for this PR? If not, let's merge it and continue the discussion on `VersionRange` and semver fun in Discord and open a follow-up if we settled on something.

skaldarnar (approved):

keturn (approved):

gestalt-module/src/main/java/org/terasology/gestalt/module/dependencyresolution/DependencyInfo.java
```diff
@@ -109,25 +108,10 @@ public void setOptional(boolean optional) {
     }
 
     /**
-     * @return The range of supported versions
+     * Returns a predicate that yields true when applied to version that is within the version range described by this dependency information
      */
-    public VersionRange versionRange() {
```
skaldarnar, line 114: @keturn this is actually used in the Terasology modules builds :see_no_evil: Guess we need to bring this method back to make that work again.

PR #119, opened 2021-06-09
feature(gradle-test): provide ablity for indexes provides by third party lib.

DarkWeird:
### Description

This PR resolves problem with adding `TeraNUI` v3.
You should not add gestalt v8 deps for TeraNUI's core module. only for nui-gestalt-di lib.

### Usage:
```java


/**
 * Index redirects for dependents libs/games.
 * Requires gestalt-di's annotation processing for dependents builds.
 */

@IndexInherited
@BindAnnotationFor(UIWidget.class)
abstract class UIWidgetIndex {
}

@IndexInherited
@BindAnnotationFor(UILayout.class)
abstract class UILayoutIndex {

}

@IndexInherited
@BindAnnotationFor(CopyStrategy.class)
abstract class CopyStrategyIndex {

}

```

### Limitations:

1. This works for any dependents libs with gestalt-di dep. not dependecies. (there can be resolved with Index task)


pollend (dismissed):

soloturn:
@BenjaminAmos is this still necessary?

BenjaminAmos:
The original intent for this pull request is obsolete since it was decided that `gestalt-inject` was minimal enough at the time to make it a dependency of NUI 4.0 (unreleased). Being able to define indexes on behalf of a third-party library may still be useful functionality though.

soloturn:
Rebased onto current develop (was conflicting). One real conflict in ClassIndexProcessor.java - develop had since switched to getBinaryName() instead of erasure().toString() for nested-class correctness; kept that fix and layered this PR's boundAnnotations.containsEntry(...) condition on top, matching the sibling method (processSubtypeIndexInReverseWay) that already had the same combination. Build passes.

@DarkWeird heads up: this force-pushed your branch, so you'll want to reset your local copy before pushing again - `git fetch origin`, then `git reset --hard origin/feature/third-party-index`.

PR #121, opened 2021-07-06
Process sub-classes when indexing & fix ClassIndex directory load path

BenjaminAmos:
### Description
This is needed for module loading and working abilities with MovingBlocks/DestinationSol#586.

Destination Sol abilities use config sub-classes (convention-based), which were not being indexed previously. The code should now index all sub-classes recursively for sub-type indexes.

When developing modules, it is common to use the directory module format with a source workspace. This was broken though, due to the use of an absolute URL path over a relative one.
> If the spec's path component begins with a slash character "/" then the path is treated as absolute and the spec path replaces the context path. 
> https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#URL(java.net.URL,%20java.lang.String)
### Testing
- If directory modules work in Destination Sol, then this works
- If abilities work in Destination Sol, then this works

### DIFF
--- a/gestalt-di/src/main/java/org/terasology/gestalt/di/index/UrlClassIndex.java
+++ b/gestalt-di/src/main/java/org/terasology/gestalt/di/index/UrlClassIndex.java
@@ -34,7 +34,7 @@ public class UrlClassIndex implements ClassIndex {
 
     public static ClassIndex byDirectory(File file) {
         try {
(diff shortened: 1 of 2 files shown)


immortius (approved):
This looks good. It might be nice to have some sort of test suite to go with this project in the future - you'ld probably need some example projects to build and interrogate, similar to how gestalt-module does its tests.

PR #122, opened 2021-08-10
refactor!: rename `Component::copy` to `Component::copyFrom`

jdrueckert:
As discussed in https://discord.com/channels/270264625419911192/869675280980115486/869683892154687498 the idea is to reduce the name clash/overlap between gestalt's `Component::copy` and `copy()` on the consuming side. Thus we can have `copy()` in other places without the concern of confusing everyone 

Quote @skaldarnar 
> One reason why I like copy-constructors is because they allow for immutable objects, and their semantic is pretty well defined.
>
> For copy methods there's always the question how they work. These are a few variants on how to make a a copy of b:
>
> - 1) a.copy(b)  may set the content of a to be a copy of b
> - 2) a = b.copy() creates a new instance as copy of b and assigns it to a
> - 3) a = Clazz.copy(b) static method to create a copy of b - bascially equivalent to a = new Clazz(b)
>
> With having JOML in the code base, there's the question whether a.copy(b) changes the state of a or b, i.e., by treating the method parameter as destination .
> The concern about correctness w.r.t. to deep copy is the same for copy constructors and copy methods - you can fail to do a deep copy correctly in both variants.
> We could leave it to the caller completely, but that's exactly the point I want to avoid :see_no_evil:

BREAKING CHANGES To fix incompatibilities simply rename implementations of `Component::copy` to `Component::copyFrom`.

### DIFF
--- a/gestalt-entity-system/src/main/java/org/terasology/gestalt/entitysystem/component/Component.java
+++ b/gestalt-entity-system/src/main/java/org/terasology/gestalt/entitysystem/component/Component.java
@@ -37,5 +37,5 @@ public interface Component<T extends Component> {
      * Copies the values from another component. This is expected to be of the same type.
      * @param other The component to copy
      */
-    void copy(T other);
+    void copyFrom(T other);
 }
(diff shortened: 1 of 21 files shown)


skaldarnar (approved):
As this is targeting the v8+ version of _gestalt_ I think we can safely merge this :+1:

PR #124, opened 2021-10-09
[release/v7.x] feat(android): add new AssetManager-based modules source

BenjaminAmos:
This pull request adds a new `ModuleFileSource` class and a derivative `ModulePathScanner` for use specifically within Android APKs (or asset bundles). Previously, it was necessary to copy all modules onto the device's filesystem before they could be loaded. This could be considered a waste of space and is not needed for other platforms. With the new `AndroidModulePathScanner` class, the only files copied onto the device's filesystem are the code dexes, which cannot be run directly from within an APK, as far as I am aware.

The new classes only work when all modules in the `assets` folder are stored in the directory module format. Jar/Zip/APK archives will not work.

### DIFF
--- a/gestalt-android/src/main/java/org/terasology/gestalt/android/AndroidAssetsFileSource.java
+++ b/gestalt-android/src/main/java/org/terasology/gestalt/android/AndroidAssetsFileSource.java
@@ -1,0 +1,170 @@
+/*
+ * Copyright 2021 The Terasology Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


Cervator (approved):
Tested out locally alongside the DS Android changes, seems to work fine! Did a local publish with Android enabled so we can have the Gestalt Android piece published to Artifactory. Am thinking with the new agent we can build that in Jenkins as well!

Merging, a complete snapshot has already been published though. We might want to try for the Android build prep then release v7.2.0

PR #125, opened 2021-10-24
[release/v7.x] fix: remove only use of Android API 24 incompatible code

### LINKED ISSUE #81: Use of ThreadLocal.withInitial prevents running on Android API 24
As far as I understand it, the minimum version of Android that gestalt supports is API 24 (Android 7.0). This is because gestalt uses certain Java 8 APIs, such as Optional, which are not available under any earlier versions.
https://github.com/MovingBlocks/gestalt/blob/4894aad4516d5264e837442ef2390ed81990f3a1/gestalt-android/build.gradle#L23-L25
At the moment though, gestalt will not run on API 24 due to the use of a method that was introduced in API 26 (Android 8.0): `java.lang.ThreadLocal.withInitial` (see the [Android documentation](https://developer.android.com/reference/java/lang/ThreadLocal?hl=en#withInitial(java.util.function.Supplier%3C?%20extends%20S%3E))). This causes the following runtime error to occur:
```
java.lang.NoSuchMethodError: No static method withInitial(Ljava/util/function/Supplier;)Ljava/lang/ThreadLocal; in class Ljava/lang/ThreadLocal; or its super classes (declaration of 'java.lang.ThreadLocal' appears in /system/framework/core-oj.jar)
    at org.terasology.gestalt.assets.management.ContextManager.<clinit>(ContextManager.java:36)
    at org.terasology.gestalt.assets.management.ContextManager.getCurrentContext(ContextManager.java:45)
    at org.terasology.gestalt.assets.management.AssetManager.resolve(AssetManager.java:151)
```
The only use of this method that I could find was in `ContextManager.java`:
https://github.com/MovingBlocks/gestalt/blob/4894aad4516d5264e837442ef2390ed81990f3a1/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/management/ContextManager.java#L36
This method is used for syntactic convenience and alternative approaches for this are already in use in `PrefabJsonFormat`:
https://github.com/MovingBlocks/gestalt/blob/4894aad4516d5264e837442ef2390ed81990f3a1/gestalt-entity-system/src/main/java/org/terasology/gestalt/entitysystem/prefab/PrefabJsonFormat.java#L72-L77

The Java implementation of this (from the OpenJDK source) behaves equivalently:
https://github.com/openjdk/jdk/blob/f613d43f269edd6f213629f45f643d0ae620b8ee/src/java.base/share/classes/java/lang/ThreadLocal.java#L142-L144
https://github.com/openjdk/jdk/blob/f613d43f269edd6f213629f45f643d0ae620b8ee/src/java.base/share/classes/java/lang/ThreadLocal.java#L291-L307

Would it be possible to change the usage in `ContextManager` to the form used in `PrefabJsonFormat`?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/92085938-use-of-threadlocal-withinitial-prevents-running-on-android-api-24?utm_campaign=plugin&utm_content=tracker%2F12992746&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F12992746&utm_medium=issues&utm_source=github).
</bountysource-plugin>

BenjaminAmos:
I have been applying this patch for a while now to get Destination Sol working on my Android testing device, which runs Android 7.0 (API 24). It simply swaps an unavaliable convenience method for its equivalent.

This fixes #81.

### DIFF
--- a/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/management/ContextManager.java
+++ b/gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/management/ContextManager.java
@@ -33,7 +33,12 @@ import java.util.Deque;
 public final class ContextManager {
 
     private static final Logger logger = LoggerFactory.getLogger(ContextManager.class);
-    private static final ThreadLocal<Deque<Name>> CONTEXT_STACK = ThreadLocal.withInitial(Queues::newArrayDeque);
+    private static final ThreadLocal<Deque<Name>> CONTEXT_STACK = new ThreadLocal<Deque<Name>>() {
+        @Override
+        protected Deque<Name> initialValue() {
+            return Queues.newArrayDeque();
+        }
+    };
 
     private ContextManager() {
     }


keturn (approved):
Oh, I see, that isn't available until API 28 (Android 8).

Having only just upgraded from Android 7.x a month ago myself, I understand.

PR #128, opened 2021-11-06
ci: add android builds to CI

### LINKED ISSUE #127: Include Android build piece in CI
With a working Android agent in Jenkins that's successfully building DestSol we should be able to apply the same approach to Gestalt to finally have its Android component built in CI and published to Artifactory. So far that's been done via local publish by a dev with a working Android Studio and a credential for Artifactory.

https://github.com/BenjaminAmos/DestinationSol/pull/1 held the initial effort for Android in DS, since then merged into the [mainline repo](https://github.com/MovingBlocks/DestinationSol), the [Android facade](https://github.com/MovingBlocks/DestSolAndroid), and our [infra repo](https://github.com/MovingBlocks/InfraPlayground)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/109732253-include-android-build-piece-in-ci?utm_campaign=plugin&utm_content=tracker%2F12992746&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F12992746&utm_medium=issues&utm_source=github).
</bountysource-plugin>

BenjaminAmos:
This is just a small change that updates the `Jenkinsfile` to use the Android build agent and enables building/publishing the `gestalt-android` library. I also fixed a compilation error in `gestalt=android-testbed` that had been previously overlooked when re-naming `Component#copy` to `Component#copyFrom`.

This should fix #127.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,10 +1,11 @@
 pipeline {
     agent {
-        label "light"
+        label "android"
     }
     stages {
         stage('Build') {
             steps {
+                sh 'echo sdk.dir=/opt/android-sdk > local.properties'
                 sh './gradlew --info --console=plain jar'
             }
         }


keturn:
Is it okay for this android agent to also publish the libraries that Java apps will use, or do we need to build those in a separate environment?

BenjaminAmos:
The android agent is just a standard java agent with the android SDK pre-installed. I don't think that the android gradle plugin is used unless gradle is directly building an android project. I often leave the android `local.properties` file in when developing and testing and it hasn't adversely affected things before. I can put the standard java modules and android modules into seperate build steps like Destination Sol does though (see [here](https://github.com/MovingBlocks/DestinationSol/blob/develop/Jenkinsfile#L7-L25)), which would mean that only the android modules are published from an android-enabled build.

soloturn:
@BenjaminAmos you want to change the target branch to develop?

BenjaminAmos:
I have rebased this pull request onto the `develop` branch now.

soloturn (approved):

PR #129, opened 2021-11-14
feat: better handle inheritance

pollend:
- include super class properties to inject into
- don't instantiate interfaces/abstract classes

### DIFF
--- a/gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/BeanDefinitionProcessor.java
+++ b/gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/BeanDefinitionProcessor.java
@@ -266,6 +266,7 @@ public class BeanDefinitionProcessor extends AbstractProcessor {
             ClassName targetClass = ClassName.get(target);
 
             // abstract/interface classes can't be instantiated
+
             if(target.getModifiers().contains(Modifier.ABSTRACT) || target.getKind() == ElementKind.INTERFACE) {
                 return builder.add("return Optional.empty();").build();
             }
@@ -316,7 +317,11 @@ public class BeanDefinitionProcessor extends AbstractProcessor {
             CodeBlock.Builder builder = CodeBlock.builder();
 
             TypeElement targetElement = target;
-            while (true) {
+            TypeMirror mirror = null;
+            do {
+                if(mirror != null) {
+                    targetElement = (TypeElement) typeUtils.asElement(mirror);
+                }
 
                 for (Element ele : targetElement.getEnclosedElements()) {
                     if (ele.getAnnotationMirrors().stream().anyMatch(in -> in.getAnnotationType().toString().equals(Inject.class.getName()))) {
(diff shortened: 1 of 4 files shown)


gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/BeanDefinitionProcessor.java
```diff
@@ -304,23 +310,34 @@ private CodeBlock buildResolution(String method, VariableElement element, ClassN
...
         private CodeBlock buildInjectionBlock(TypeElement target) {
             String name = "instance";
             CodeBlock.Builder builder = CodeBlock.builder();
 
-            for (Element ele : target.getEnclosedElements()) {
-                if (ele.getAnnotationMirrors().stream().anyMatch(in -> in.getAnnotationType().toString().equals(Inject.class.getName()))) {
-                    if (ele instanceof ExecutableElement) {
-                        List<? extends VariableElement> parameters = ((ExecutableElement) ele).getParameters();
+            TypeElement targetElement = target;
+            while (true) {
```
DarkWeird: Don't use `while (true)` please.
I see how this works... but this line terrifies me.

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/BeanDefinitionProcessor.java
```diff
@@ -264,6 +265,11 @@ private CodeBlock buildObjectPointCreationBlock(TypeElement target) {
             CodeBlock.Builder builder = CodeBlock.builder();
             ClassName targetClass = ClassName.get(target);
 
+            // abstract/interface classes can't be instantiated
+            if(target.getModifiers().contains(Modifier.ABSTRACT) || target.getKind() == ElementKind.INTERFACE) {
+                return builder.add("return Optional.empty();").build();
```
DarkWeird, line 271: How should works this line?
`MyAbstractImplementation` looks like usual non-abstract class..

DarkWeird (changes requested):
This PR breaking getting beans by interfaces and abstracts.
1. `CollectionResolveTest` is broken
2.  modified copy of `AbstractBeanTest#injectAbstractBean`:
```

    @Test
    public void injectAbstractBeanGetAbstract() {
        ServiceRegistry registry = new ServiceRegistry();
        registry.with(AbstractBeanTest.Impl1.class).lifetime(Lifetime.Singleton);
        registry.with(AbstractBeanTest.Impl2.class).lifetime(Lifetime.Singleton);
        registry.with(ConcreteImplementationFromAbs.class).lifetime(Lifetime.Singleton);

        BeanContext cntx = new DefaultBeanContext(registry);
        MyAbstractImplementation c1 = cntx.getBean(MyAbstractImplementation.class);

        Assert.assertNotNull(c1.impl);

    }
```
this modified test for getting as super class. don't works.

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/BeanDefinitionProcessor.java
pollend: some classes can't be instantiated. I guess the logic is faulty. :/

DarkWeird (approved):

PR #130, opened 2021-11-17
feat: ensure bean definition class name is unique

pollend:
ensure the class name is unique. `EnclosedName$$staticClass`


pollend:
darn looks like there is some logic that I'm missing :/

DarkWeird (changes requested):
Use one `$` please T_T

With `$`:
BeanDefinition became like nested class(of nested class), and it will same for nested and not nested classes.
(for The True Nested class we should describe it in BeanClass... but it is anitpattern for annotation processing)


With `$$`:
Class with strange name `Parent$$Nested`, which have 'nested' BeanDefinition..

DarkWeird (approved):

PR #131, opened 2021-11-25
[develop] feat(android): add new AssetManager-based modules source

BenjaminAmos:
This a variant of #124 adapated for the `develop` branch of gestalt. It addtionally adds the ability to load module dexes from `.jar` files, rather than just as a series of raw `.dex` files in a directory. This is needed because Android requires `jar` file sources for modules to access classpath resources, which `gestalt-di` uses.


gestalt-module/src/main/java/org/terasology/gestalt/module/ModuleServiceRegistry.java
```diff
@@ -6,9 +6,7 @@
 
 public class ModuleServiceRegistry extends ServiceRegistry {
     public ModuleServiceRegistry(PermissionProviderFactory permissionProviderFactory) {
-        with(ModulePathScanner.class);
         with(ModuleMetadataJsonAdapter.class);
-        with(ModulePathScanner.class);
```
BenjaminAmos, line 11: I removed `ModulePathScanner` from the default context because on Android you might want to use `AndroidModulePathScanner` instead. This would not be possible if the concrete base class is always provided.

pollend (approved):

PR #136, opened 2022-06-07
build: upgrade gradle to 6.9.2 (from 6.8.2)

### LINKED ISSUE #116: upgrade to gradle 6.9.x (prep for gradle 7.x)
Gradle 7 has some features Terasology is interested in (e.g. version catalogs). Since the MovingBlocks/✳ libraries are often used in composite builds with Terasology, it seems like a good idea to keep their gradle versions from drifting too far apart.

I propose upgrading all the libraries to the latest in the 6.x line (currently 6.9.0) and fixing whatever deprecation warnings crop up there. That should give us a smooth path to then adopting Gradle 7.

keturn:
in preparation for upgrading MovingBlocks builds to gradle v7.

gradle scan didn't flag anything as deprecated here. Looks like this project will be good for v7!

Resolves #116


soloturn:
tried with gradle-8.2 and it works, see #137

jdrueckert:
Superseded by #137

PR #145, opened 2024-07-28
https://artifactory.terasology.io is now

soloturn:


BenjaminAmos (changes requested):
This is almost right but you missed one important instance of the old URL:
https://github.com/MovingBlocks/gestalt/blob/c706ad097378631f40d43cfd5b6633ed61c3f73e/gradle/common.gradle#L71

BenjaminAmos (approved):

PR #146, opened 2024-07-28
gradle-8.9 to permit java-22

soloturn:
run
``` 
gradle wrapper --gradle-version 8.9
``` 
and leave the default values generated in place, like the neetwork timeouts.

rootProject.buildDir is deprecated. as `gradle clean` anway remove builddir by default, delete it from build.gradle. switch to -bin distribution as it is smaller. all is only usefull if you work on the build, with groovy, which is rare. 

see documentation here:
https://youtrack.jetbrains.com/issue/IDEA-203597

especially:
When using the Kotlin DSL and opening a .gradle.kts script for the first time, the IDE resolver will download the -src distro and use the sources from it automatically.


BenjaminAmos:
Would it be possible to split the conversion of the gradle build to Kotlin into a separate pull request? It is far easier to approve upgrading gradle first, since that is far less likely to cause build regressions.

(This is not a hard requirement, just a personal preference.)

soloturn:
> Would it be possible to split the conversion of the gradle build to Kotlin into a separate pull request? It is far easier to approve upgrading gradle first, since that is far less likely to cause build regressions.
> 
> (This is not a hard requirement, just a personal preference.)

split @BenjaminAmos

jdrueckert (commented):
This change seems fine content-wise, but I'm missing change documentation, a.k.a some info on why certain changes were made.

In particular removing the `clean` task (which is kind of given in the commit message of the second commit, so copying that in the PR description should be easy enough) and using the `-bin` package instead of the `-all` package like we did before.
Also, where does the value for the network timeout come from? Is that some kind of default / best practice / dice roll?

soloturn:
updated description @jdrueckert

jdrueckert (approved):
lgtm

PR #147, opened 2024-07-28
Prepare kotlin, version catalog

soloturn:
this introduces a version catalog in settings.gradle.kts, as described here, to specify the version once, removing the ancient way for it.
https://docs.gradle.org/current/userguide/platforms.html

otherwise just syntactic stuff, necessary for kotlin and not for groovy.


BenjaminAmos (approved):
This seems like a general improvement.

PR #152, opened 2024-11-03
either this file is generated and ignored, or commited. but not both.

soloturn:
remove jar file for testing module load, it is generated.


BenjaminAmos (changes requested):
The gitignore changes are fine but could you please remove your changes to the file `gestalt-es-perf/test-modules/moduleF.jar`?

soloturn:
> The gitignore changes are fine but could you please remove your changes to the file `gestalt-es-perf/test-modules/moduleF.jar`?

i cannot follow? the PR does not change moduleF.jar, but deletes it. the purpose of gitignore is to make sure untracked files are not tracked. if the file is in the repo gitignore does nothing any more. gradle itself cannot yet produce byte-for-byte similar jar files.

references:
*  https://stackoverflow.com/questions/45400361/why-is-gitignore-not-ignoring-my-files
* https://github.com/gradle/gradle/issues/30871

BenjaminAmos:
You are correct and I misunderstood. GitHub was presenting the diff very misleadingly, so I thought you had shrunk the file rather than deleting it.

BenjaminAmos (approved):

PR #158, opened 2025-01-19
javadoc for gestalt-inject-java

soloturn:


gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/AnnotationTypeWriter.java
```diff
@@ -12,22 +12,38 @@
 import java.util.HashSet;
 import java.util.Map;
 
+/**
+ * Class to write annotation types.
```
BenjaminAmos: The `AnnotationTypeWriter` class is responsible for writing to files under `META-INF/annotations` listing which classes are annotated with a particular annotation. For example, types annotated with `@IndexInherited` would be listed in `META-INF/annotations/org.terasology.context.annotation.IndexInherited`.

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/AnnotationTypeWriter.java
```diff
@@ -12,22 +12,38 @@
...
+ */
 public class AnnotationTypeWriter {
     private static final String META_INF = "META-INF/annotations";
 
     private final Filer filer;
     private final Map<String, HashSet<String>> results = new HashMap<>();
     private final Map<String, FileObject> files = new HashMap<>();
 
+    /**
+     * Creates a AnnotationTypeWriter to write an annotation.
```
BenjaminAmos:
```suggestion
     * Creates an AnnotationTypeWriter.
```

Short and not very useful but this constructor doesn't do much anyway. This is probably enough.

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/AnnotationTypeWriter.java
```diff
@@ -12,22 +12,38 @@
...
+     * @param filer support creating files by annotation processor.
+     */
     public AnnotationTypeWriter(Filer filer) {
         this.filer = filer;
     }
 
+    /**
+     * Writes an annotation to a service.
+     * @param service the service to write.
+     * @param target the target.
```
BenjaminAmos: This does not actually do any writing to files yet, so I will admit that the name is misleading. What it does is register that a given class should be associated as being annotated by a given annotation (aka. service).

In this instance, a service refers to a particular annotation type. So, the service parameter would be the fully-qualified type name of the annotation.

The target is the target class that was annotated with that annotation.

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/AnnotationTypeWriter.java
```diff
@@ -12,22 +12,38 @@
...
+     */
     public void writeAnnotation(String service, String target) {
         results.putIfAbsent(service, new HashSet<>());
         results.get(service).add(target);
     }
 
+    /**
+     * Writes class to file.
```
BenjaminAmos:
```suggestion
     * Writes out annotation-class mappings to files.
```

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/BeanDefinitionProcessor.java
```diff
@@ -66,13 +69,23 @@ public class BeanDefinitionProcessor extends AbstractProcessor {
     };
     private final Set<String> processed = new HashSet<>();
     private final Set<String> beanDefinitions = new HashSet<>();
+    /**
+     * the elements.
```
BenjaminAmos: `Elements` refers to [`javax.lang.model.util.Elements`](https://docs.oracle.com/en/java/javase/17/docs/api/java.compiler/javax/lang/model/util/Elements.html). It is a set of utilities for manipulating Java syntax elements, not a collection of elements, as the name might misleadingly suggest.

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/BeanDefinitionProcessor.java
```diff
@@ -66,13 +69,23 @@ public class BeanDefinitionProcessor extends AbstractProcessor {
...
+     */
     protected Elements elementUtils;
+    /**
+     * the types.
```
BenjaminAmos: `Types` refers to [`javax.lang.model.util.Types`](https://docs.oracle.com/en/java/javase/17/docs/api/java.compiler/javax/lang/model/util/Types.html). It is a set of utilities for manipulating Java types, not a collection of types, as the name might misleadingly suggest.

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/BeanDefinitionProcessor.java
```diff
@@ -66,13 +69,23 @@ public class BeanDefinitionProcessor extends AbstractProcessor {
...
+     */
     protected Types typeUtils;
     private Filer filer;
     private ServiceTypeWriter writer;
     private ElementUtility utility;
     private Messager messager;
 
+    /**
+     * Init proocessingEnvironment
+     * @param processingEnvironment the ProcessingEnvironment.
```
BenjaminAmos:
```suggestion
     * {@inheritDoc}
```

This is an override from an already well-documented base method, so we might as well inherit that.

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/BeanDefinitionProcessor.java
```diff
@@ -86,11 +99,19 @@ public synchronized void init(ProcessingEnvironment processingEnvironment) {
         this.utility = new ElementUtility(processingEnvironment.getElementUtils(), processingEnvironment.getTypeUtils());
     }
 
+    /**
+     * {@return the supported annotation types.}
```
BenjaminAmos:
```suggestion
     * {@inheritDoc}
```
This is an override from an already well-documented base method, so we might as well inherit that.

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/BeanDefinitionProcessor.java
```diff
@@ -86,11 +99,19 @@ public synchronized void init(ProcessingEnvironment processingEnvironment) {
...
+     */
     @Override
     public Set<String> getSupportedAnnotationTypes() {
         return Collections.singleton("*");
     }
 
+    /**
+     * {@return true if bean definitions could be processes, fals otherwise.}
+     * @param annotations the annotations.
+     * @param environment the environment.
```
BenjaminAmos:
```suggestion
     * {@inheritDoc}
```
This is an override from an already well-documented base method, so we might as well inherit that.

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/ResourceProcessor.java
```diff
@@ -20,6 +20,9 @@
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+/**
+ * Annotation processor to create resources for types.
```
BenjaminAmos:
```suggestion
 * Annotation processor to list resources for types.
```

This processor does not create the resources, it just lists the resources that are there. We need to do this because listing resources packed in an APK on Android is prohibitively expensive.

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/ElementUtility.java
```diff
@@ -11,23 +11,41 @@
 import javax.lang.model.util.Types;
 import java.util.List;
 
+/**
+ * BenjaminAmos, pls proper description.
```
BenjaminAmos:
```suggestion
 * Utilities for discovering and distinguishing element types.
```

This is a utility class: convenient but just a mixture of related useful routines. Most projects have one somewhere, even if it's not considered the best programming practice sometimes.

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/ClassIndexProcessor.java
```diff
@@ -24,6 +24,9 @@
 import java.util.Queue;
 import java.util.Set;
 
+/**
+ * BenjaminAmos pls description which makes sense.
```
BenjaminAmos:
```suggestion
 * Records all types annotated with `@Index` or `@IndexInherited`, as well as all descendants of a type annotated with `@IndexInherited`.
```

BenjaminAmos (commented):
This is a first-pass review. I have not looked over everything yet and will carry-on another time. It should be enough to start with.

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/BeanDefinitionProcessor.java
```diff
@@ -154,9 +172,21 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
         return false;
     }
 
+    /**
+     * Class that writes the definition.
+     */
     public class DefinitionWriter extends ElementScanner8<Object, String> {
+        /**
+         * the base package.
```
BenjaminAmos:
```suggestion
         * The base package containing gestalt annotations.
```

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/BeanDefinitionProcessor.java
```diff
@@ -154,9 +172,21 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
...
+         */
         public static final String BASE_PACKAGE = "org.terasology.context";
+        /**
+         * arguments.
```
BenjaminAmos:
```suggestion
         * Argument field placeholder.
```

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/BeanDefinitionProcessor.java
```diff
@@ -154,9 +172,21 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
...
+         */
         public static final String ARGUMENT_FIELD = "$ARGUMENT";
+        /**
+         * class metadata.
```
BenjaminAmos:
```suggestion
         * Class metadata field placeholder.
```

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/ElementUtility.java
```diff
@@ -40,7 +58,11 @@ public final TypeElement classElementFor(Element element) {
         return null;
     }
 
-
+    /**
+     * {@return true, if stereotype contains element, false otherwise.}
+     * @param element the element.
+     * @param stereotype the steriotype.
```
BenjaminAmos:
```suggestion
     * @param stereotype the stereotype.
```

gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/ServiceTypeWriter.java
```diff
@@ -12,22 +12,37 @@
...
+    /**
+     * Creates a ServiceTypeWriter to write a subtype to a file.
+     * @param filer support creating files by annotation processor.
+     */
     public ServiceTypeWriter(Filer filer) {
         this.filer = filer;
     }
 
+    /**
+     * Writes a srvice to a file.
```
BenjaminAmos:
```suggestion
     * Writes a service to a file.
```

BenjaminAmos (commented):
Here are a few other obvious things I noticed. Apologies for the long delay!

BenjaminAmos (approved):

Issue #134, opened 2022-05-27
update dependencies


keturn:
Now that [IntelliJ checks dependencies against a vulnerability database](https://blog.jetbrains.com/idea/2022/04/ensure-greater-software-security-with-package-analysis-by-checkmarx-in-intellij-idea/), we see gestalt pulling in some dependencies with known CVEs. Either directly or transitively through things like Reflections.

We should update the dependencies or add [constraints](https://docs.gradle.org/current/userguide/dependency_constraints.html#sec:adding-constraints-transitive-deps) to the transitive dependencies, as appropriate.

Related to MovingBlocks/Terasology#5021

soloturn:
Still relevant - a concrete current instance: CodeRabbit's review on #170 (dependency locking) flagged `commons-vfs2:2.2` for CVE-2025-27553 and CVE-2025-30474, declared directly in `gestalt-module/build.gradle.kts` and `gestalt-inject-java/build.gradle.kts`, still present in the lockfiles across the repo including testpack/*. No tracked source actually uses the VFS API - removing the dependency (or upgrading to 2.10.0+) and regenerating the lockfiles would close this out.

soloturn:
Closing - fixed by #171 (removes the unused commons-vfs2 dependency entirely).

PR #159, opened 2025-01-26
build: Added configurations to apply constraints to dependencies `resolves` #134

Thedrogon:
Added configurations to apply constraints to the `build.gradle` dependencies to prevent  `gestalt` pulling in dependencies with known CVEs, directly or transitively through things like Reflections. Added some suppressions in various files and as well removed some unnecessary imports

### DIFF
--- a/gestalt-android-testbed/build.gradle
+++ b/gestalt-android-testbed/build.gradle
@@ -1,8 +1,8 @@
 // Copyright 2021 The Terasology Foundation
 // SPDX-License-Identifier: Apache-2.0
 plugins {
-    id("com.android.application")
-    id("project-report")
+    id 'com.android.application'
+    id 'project-report'
 }
 
 android {
@@ -16,8 +16,8 @@ android {
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
     }
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility = JavaVersion.VERSION_1_8
+        targetCompatibility = JavaVersion.VERSION_1_8
     }
     packagingOptions {
         merge "META-INF/annotations/*"
@@ -34,21 +34,36 @@ android {
 }
 
 dependencies {
-    implementation(fileTree(dir: "libs", include: ["*.jar"]))
-    implementation("com.android.support:appcompat-v7:28.0.0")
-    implementation("com.android.support.constraint:constraint-layout:1.1.3")
-    implementation(project(":gestalt-android"))
-    implementation(project(":gestalt-asset-core"))
-    implementation(project(":gestalt-entity-system"))
-    implementation(libs.guava)
-    implementation(libs.slf4j.api)
-    implementation("com.github.tony19:logback-android:1.3.0-3")
-    implementation(project(":testpack:testpack-api"))
-    testImplementation(libs.junit)
-    androidTestImplementation("com.android.support.test:runner:1.0.2")
-    androidTestImplementation("com.android.support.test.espresso:espresso-core:3.0.2")
+    implementation fileTree(dir: "libs", include: ["*.jar"])
+    implementation "com.android.support:appcompat-v7:28.0.0"
+    implementation "com.android.support.constraint:constraint-layout:1.1.3"
+    implementation project(":gestalt-android")
+    implementation project(":gestalt-asset-core")
+    implementation project(":gestalt-entity-system")
+    implementation libs.guava
+    implementation libs.slf4j.api
+    implementation "com.github.tony19:logback-android:1.3.0-3"
+    implementation project(":testpack:testpack-api")
+    testImplementation libs.junit
+    androidTestImplementation "com.android.support.test:runner:1.0.2"
+    androidTestImplementation "com.android.support.test.espresso:espresso-core:3.0.2"
 
-    annotationProcessor(project(":gestalt-inject-java"))
+    annotationProcessor project(":gestalt-inject-java")
+}
+
+configurations {
+    // Apply constraints to enforce specific versions
+    all {
+        resolutionStrategy {
+            force(
+                "com.android.support:appcompat-v7:28.0.0",
+                "com.android.support.constraint:constraint-layout:1.1.3",
+                "com.github.tony19:logback-android:1.3.0-3",
+                "com.android.support.test:runner:1.0.2",
+                "com.android.support.test.espresso:espresso-core:3.0.2"
+            )
+        }
+    }
 }
 
 // Android projects don't provide direct access to compileJava but this seems to work.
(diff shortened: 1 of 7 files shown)


gestalt-android-testbed/build.gradle
```diff
@@ -1,8 +1,8 @@
 // Copyright 2021 The Terasology Foundation
 // SPDX-License-Identifier: Apache-2.0
 plugins {
-    id("com.android.application")
-    id("project-report")
```
BenjaminAmos: For essentially all the changes in this file: this is a reason why things are like this. We are preparing to migrate these build scripts to use the Kotlin DSL eventually. The first step was adapting the files to use Kotlin-compatible syntax, which is why they might not appear as idomatic Groovy for the moment.

Please revert any changes to this file outside of the `configurations` block.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AbstractFragmentDataProducer.java
```diff
@@ -69,6 +69,7 @@ protected AbstractFragmentDataProducer(AssetManager assetManager, Class<U> rootA
     public Set<Name> getModulesProviding(Name resourceName) {
         if (resolveModuleFromRoot) {
             return ImmutableSet.copyOf(Collections2.transform(assetManager.resolve(resourceName.toString(), rootAssetType), new Function<ResourceUrn, Name>() {
+                @SuppressWarnings("null")
```
BenjaminAmos: I would rather not suppress a genuine warning.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/Asset.java
```diff
@@ -172,6 +172,7 @@ public boolean equals(Object obj) {
             return true;
         }
         if (obj instanceof Asset) {
+            @SuppressWarnings("rawtypes")
```
BenjaminAmos: Please do not suppress warnings without good justification.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -67,6 +69,8 @@
  */
 @API
 @ThreadSafe
+@SuppressWarnings({ "unchecked", "removal" })
```
BenjaminAmos: These warnings should not be suppressed without good reason.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -298,7 +300,7 @@ synchronized void registerAsset(Asset<U> asset, DisposalHook disposer) {
     }
 
     /**
-     * Creates and returns an instance of an asset, if possible. The following methods are used to create the copy, in order, with the first technique to succeeed used:
+     * Creates and returns an instance of an asset, if possible. The following methods are used to create the copy, in order, with the first technique to succeed used:
```
BenjaminAmos, line 303: This is a valid typo fix. It is probably the only valid change to this file.

BenjaminAmos (changes requested):
Thanks for this!

Constraining the dependencies is a useful adjustment but I would rather it was not bundled with other wide-sweeping changes. Suppressing warnings is a last resort as well, since usually we want to see warnings.

Thedrogon (commented):
reverted the changes .

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -45,9 +46,9 @@
 import java.lang.ref.ReferenceQueue;
 import java.lang.ref.WeakReference;
 import java.lang.reflect.Type;
-import java.security.AccessController;
-import java.security.PrivilegedActionException;
-import java.security.PrivilegedExceptionAction;
+
+//import java.security.PrivilegedActionException;  //should i even remove these???(not sure)
+//import java.security.PrivilegedExceptionAction;
```
soloturn: if not used please remove ...

soloturn:
@BenjaminAmos does this look good to you now? apart from commented stuff which imo should be better removed ...

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
Thedrogon: Removed them.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -328,16 +330,16 @@ Optional<T> createInstance(Asset<U> asset) {
         Optional<? extends Asset<U>> result = asset.createCopy(asset.getUrn().getInstanceUrn());
         if (!result.isPresent()) {
             try {
-                return AccessController.doPrivileged((PrivilegedExceptionAction<Optional<T>>) () -> {
```
BenjaminAmos, line 331: Despite being deprecated, I'm afraid that for the time being we do still need to use the `AccessController` APIs. Gestalt sandboxes modules using `SecurityManager` as an additional precaution and until it's completely unusable I'd rather have that safety net than not. The `AccessController.doPrivileged` calls are needed when running with a `SecurityManager` active.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
```diff
@@ -131,7 +133,7 @@ public synchronized void close() {
     /**
      * Disposes any assets queued for disposal. This occurs if an asset is no longer referenced by anything.
      */
-    @SuppressWarnings("unchecked")
+    //@SuppressWarnings("unchecked")
```
BenjaminAmos, line 136: You can either remove those suppressions or leave them be but commenting them out like that in committed code is rather messy.

BenjaminAmos (changes requested):
I appreciate you trying to make changes for the better but not all warnings are actionable here for reasons that I have tried to explain.

If you could constrain the changes to that tiny but beneficial gradle constraint improvement then this would be much easier to take in. Removing logic in the code without understanding why it is there (this library is quite a central piece of our games) just makes this pull request unviable as-is, I'm afraid.

gestalt-asset-core/src/main/java/org/terasology/gestalt/assets/AssetType.java
Thedrogon: Then for the requirement I will revert them in the PR.

soloturn:
@BenjaminAmos what you think now of this?

BenjaminAmos:
My comments from before are still valid, I think? The changes remove code needed for gestalt to work with `SecurityManager`, as well as other unrelated changes. I could probably accept removing unused code but the `AccessController.doPrivileged` blocks are still needed.

BenjaminAmos:
This also has merge conflicts now, so it would have to be updated regardless.

PR #162, opened 2026-07-26
fix(gestalt-inject-java): use binary name for nested classes in class index

soloturn:
## Summary
- `ClassIndexProcessor` wrote nested types (e.g. `Outer.Inner`) using `TypeMirror#toString()`, which renders the dotted/canonical form instead of the `Outer$Inner` binary name `ClassLoader.loadClass()` requires. This made `APIScanner.scan()` throw `ClassNotFoundException` for any `@API`-annotated nested class or interface (surfaced via Terasology's `MeshBuilder.TextureMapper`). Fixed both `processAnnotationIndex` and `processSubtypeIndexByDirectMarked` to use `Elements#getBinaryName()`, matching the pattern already used by `processSubtypeIndexInReverseWay`.
- Bumped the Gradle wrapper from 8.10.2 to 9.6.1 since 8.10.2 doesn't run on newer JDKs (e.g. 26).

## Test plan
- [x] `gradle :gestalt-inject-java:compileJava` succeeds
- [ ] Downstream Terasology `Missing facet provider` / `ClassNotFoundException` symptoms resolved once this is republished

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/ClassIndexProcessor.java
+++ b/gestalt-inject-java/src/main/java/org/terasology/gestalt/annotation/processing/ClassIndexProcessor.java
@@ -97,8 +97,12 @@ public class ClassIndexProcessor extends AbstractProcessor {
                         TypeMirror candidate = supers.poll();
                         if (candidate.getKind() != TypeKind.NONE) {
                             if (elementUtility.hasStereotype(elementUtility.getTypes().asElement(candidate),
-                                    Collections.singletonList(IndexInherited.class.getName())))
-                                subtypesTypeWriter.writeSubType(elementUtility.getTypes().erasure(candidate).toString(), elementUtility.getTypes().erasure(type.asType()).toString());
+                                    Collections.singletonList(IndexInherited.class.getName()))) {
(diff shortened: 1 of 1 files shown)


BenjaminAmos (approved):
This fixes a very large oversight - thank you!

There do not appear to be many unit tests covering the `@Index` and `@IndexInherited` annotations, so this behaviour was not previously validated. In future, it would be good to add a unit test to document the correct intended behaviour.

soloturn:
how is this released and used in terasology, @BenjaminAmos ?

BenjaminAmos:
In theory, a snapshot should have been published automatically, which Terasology should pull in because we have not been very good at tagging a release of Gestalt 8 to date. One of the artifacts was published [here](https://artifactory.terasology.io/artifactory/libs-snapshot-local/org/terasology/gestalt/gestalt-util/8.0.0-SNAPSHOT/gestalt-util-8.0.0-20260726.081650-28.jar), for example. In reality, somehow Terasology depends on `gestalt-8.0.1-SNAPSHOT` (introduced in https://github.com/MovingBlocks/Terasology/commit/0d310c8c3463feaa536b6d6b7fd79cd5f73ad8d4), which appears to have been published manually at some point, so @Cervator would probably be able to explain why things are in an inconsistent state right now. Based on the commit message, I suspect that version 8.0.1 is actually #160 right now, which should have already been merged but was overlooked.

PR #165, opened 2026-07-29
ci: add Android Java API compatibility checks

BenjaminAmos:
This pull request introduces the [`gradle-animalsniffer-plugin`](https://github.com/xvik/gradle-animalsniffer-plugin) into the testing process to verify that Gestalt libraries do not call APIs which would be unavailable at runtime on Android. The API signatures are provided by the [`gummy-bears`](https://github.com/open-toast/gummy-bears) project.

The legacy `common.gradle.kts` shared logic is also moved into a more modern Gradle convention plugin, which should be easier to work with.

### Further Rationale
Android is an unusual platform from a build perspective due to it having lacking support for many Java APIs. For a long time, approximately Java 8 was the only supported baseline. Java 11 support was introduced in Android 12 (2021) and Java 17 support in Android 14 (2023). We still support all the way back to Android 7 currently (my preference).

Typically, if you wanted to ensure that code still ran under an older JVM runtime, you would compile the code using `-source <java_version>` and `-target <java_version>`, with `-release <java_version>` replacing these two from Java 9. This ensures that the code is compatible by limiting the available classes but also has the consequence of limiting you to using solely the language features supported by that old version.

The Android build tools offers [the `D8` dexer](https://developer.android.com/studio/write/java8-support) as an alternative. Strictly speaking, D8 is a tool used to translate JVM bytecode into a format compatible with Android's runtime. As part of this process, it also performs a "desugaring" step, allowing features from newer Java runtimes to be backported at build time in a few cases. This breaks the assumptions of the Java compiler that language features are tied to specific runtimes and allows us to use newer language features without any of the corresponding newer APIs being available. If any of those unavailable APIs happen to be in-use, then it will likely result in an unexpected runtime failure when the class cannot be found. The plugin introduced here performs more strict checks on the classes used by our libraries in an attempt to prevent such scenarios.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ hs_err_pid*
 
 # Ignore Gradle
 /.gradle/
+/build-logic/.gradle
 build/
 local.properties
 
--- a/build-logic/build.gradle.kts
+++ b/build-logic/build.gradle.kts
@@ -1,0 +1,13 @@
+// Copyright 2026 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+plugins {
+    `kotlin-dsl`
+}
+
+repositories {
+    gradlePluginPortal()
+}
+
+dependencies {
+}
--- a/build-logic/build.gradle.kts
+++ b/build-logic/build.gradle.kts
@@ -10,4 +10,5 @@ repositories {
 }
 
 dependencies {
+    implementation("ru.vyarus:gradle-animalsniffer-plugin:2.0.1")
 }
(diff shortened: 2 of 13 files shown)


Cervator:
This rings a bell, but a very vague one since it has been years since I looked at this last - but I do remember "cheaty" ways to go use some newer language features without breaking backwards compatibility. It is exciting to see that again! And once more make me think about Terasology maybe eventually making it to Android, one day :-)

Changes look fine to me, but I suppose the big question here is testing everything. That might take just a bit due to the extra steps for the setup. I think I've done it before and may even have Android Studio set up and working around her somewhere. Plus: old Android devices!

soloturn (approved):

PR #167, opened 2026-07-30
fix(module): log APIScanner class-not-found at info, not warn+stacktrace

soloturn:
> **AI-assisted change proposal.** Filed by agent driven by @soloturn via [GDD](https://siliconsaga.github.io/yggdrasil/gdd/).

## Summary

- `APIScanner.scan()` logged a full stack trace at WARN for every module class
  it couldn't resolve on the given classloader. This fires routinely — the
  engine's `setupSandbox()` scans every registered module's class index
  before that module's classes are loaded onto the classloader, so most
  non-engine modules trip it on every startup. Downgraded to a one-line INFO
  naming the class and the classloader checked.

## Test plan

- [ ] Run `Terasology Latest.app` and confirm startup no longer logs
      `ClassNotFoundException` stack traces from `APIScanner`, just INFO lines

## Related

- See jenkins.terasology.io Omega build console output that surfaced this

### DIFF
--- a/gestalt-module/src/main/java/org/terasology/gestalt/module/sandbox/APIScanner.java
+++ b/gestalt-module/src/main/java/org/terasology/gestalt/module/sandbox/APIScanner.java
@@ -78,7 +78,7 @@ public class APIScanner {
                     }
                 }
             } catch (ClassNotFoundException e) {
-                logger.info("Class {} not found on classloader {} - expected if it belongs to a module's own classloader instead", apiClass, forClassLoader);
+                logger.info("Class {} not found on classloader {}", apiClass, forClassLoader);
             }
         }
     }
--- a/gestalt-module/src/main/java/org/terasology/gestalt/module/sandbox/APIScanner.java
+++ b/gestalt-module/src/main/java/org/terasology/gestalt/module/sandbox/APIScanner.java
@@ -78,7 +78,7 @@ public class APIScanner {
                     }
                 }
             } catch (ClassNotFoundException e) {
-                logger.warn("Class not found", e);
+                logger.info("Class {} not found on classloader {} - expected if it belongs to a module's own classloader instead", apiClass, forClassLoader);
             }
         }
     }


BenjaminAmos:
> This fires routinely — the engine's setupSandbox() scans every registered module's class index
before that module's classes are loaded onto the classloader, so most
non-engine modules trip it on every startup.

This seems more like the issue, rather than the warning itself. Destination Sol gets a lot of these warnings at runtime as well. If I understand correctly, is this effectively prohibiting any modules outside of the engine classpath from being exported to other modules as APIs?

soloturn:
> **AI-assisted comment.** Filed by agent driven by @soloturn via [GDD](https://siliconsaga.github.io/yggdrasil/gdd/).

@BenjaminAmos good question - I traced it through `JavaModuleClassLoader.loadClass()`:

```java
ClassLoader parentLoader = ObtainClassloader(clazz);
if (parentLoader != this && !(parentLoader instanceof ModuleClassLoader)) {
        return clazz;
    } else {
        logger.error("Denied access...");
        return null;
    }
}
return clazz;  // module-to-module access: no check at all
```

The permission check only fires when the resolved class's own loader is **neither this module's classloader nor any other `ModuleClassLoader`** - i.e. it only gates access to engine/system-classloader classes. Any class whose loader *is* a `ModuleClassLoader` (any other module, including a dependency) returns unchecked. So module-to-module API exports aren't gated by this mechanism at all - DestinationSol's own APIs aren't blocked by it.

What the warning storm actually is: `ModuleManager.setupSandbox()` loops over every registered module and calls `apiScanner.scan(module.getClassIndex())` against the system classloader - including modules whose classes were never going to be checked against this permission set in the first place. So it's dead work, not a real error, but you're right that the scan loop is the actual issue, not just the log level.

Filed the underlying fix separately since `setupSandbox()` lives in the engine repo, not here: MovingBlocks/Terasology#5353 restricts the scan to the engine module (and dev-mode classpath modules) instead of every registered module. This PR (downgrading the log level) still stands on its own regardless - even with #5353, `ClassNotFoundException` from other causes should log at info, not warn+stacktrace, for expected-miss cases.

soloturn:
> **AI-assisted comment.** Filed by agent driven by @soloturn via [GDD](https://siliconsaga.github.io/yggdrasil/gdd/).

For anyone wondering whether this PR is now redundant with MovingBlocks/Terasology#5353: it isn't, and we're keeping both.

#5353 fixes the one known call site (`ModuleManager.setupSandbox()`) so it stops scanning modules that were never resolvable via the system classloader - the common case goes away there.

This PR fixes `APIScanner` itself, which is a general-purpose method in a shared library - `forClassLoader`/`classIndex` are caller-supplied, so any other consumer (tests, other games built on gestalt, or a future Terasology call site) can still legitimately hit a mismatch. Even within Terasology, the malformed `MeshBuilder.TextureMapper`-with-a-dot-instead-of-`$` entry visible in the original log suggests the class index can contain bad entries independent of which module is being scanned - #5353 doesn't touch that.

So this is defense-in-depth, not overlap: even after #5353 lands, any remaining `ClassNotFoundException` here is still the expected-miss case, not something that should dump a stack trace.

soloturn:
> **AI-assisted comment.** Filed by agent driven by @soloturn via [GDD](https://siliconsaga.github.io/yggdrasil/gdd/).

Rebased onto current `develop` (was 3 commits behind), force-pushed — clean, no conflicts. Ready for review.

gestalt-module/src/main/java/org/terasology/gestalt/module/sandbox/APIScanner.java
```diff
@@ -78,7 +78,7 @@ public void scan(ClassIndex classIndex) {
                     }
                 }
             } catch (ClassNotFoundException e) {
-                logger.warn("Class not found", e);
+                logger.info("Class {} not found on classloader {} - expected if it belongs to a module's own classloader instead", apiClass, forClassLoader);
```
BenjaminAmos:
```suggestion
                logger.info("Class {} not found on classloader {}", apiClass, forClassLoader);
```

It would be best not to assume the context here. It is still useful to know which class loader failed in addition though.

soloturn: Fixed in c524e4a - dropped the assumption, kept the class/classloader detail.

soloturn:
@BenjaminAmos please review

Cervator (approved):
Going to approve and merge then leave a maybe-related question on chat (maybe unrelated bit about `@API` and MTE using stuff for testing)

